PR #2, opened 2018-05-19
Laiaaa0 1

laiaaa0:
Hi! :)
I did a first attempt at implementing assignment 1

### DIFF
--- a/laiaaa0/assignment1/q1.cpp
+++ b/laiaaa0/assignment1/q1.cpp
@@ -1,0 +1,77 @@
+#include <iostream>
+#include <vector>
+using namespace std;
+
+bool anagrams_caseSensitive (string & s1, string & s2){
+  // Compare lengths. If they are not the same, return false.
+  if (s1.size() != s2.size()) return false;
+
+  // Create a vector of the size of the ascii table.
+  std::vector<int> counts(128);
+
+  // For each of the characters in the strings
+  for (int i=0; i<s1.size(); i++){
+    //Add one count to the ascii position of the current character of the first string
+    counts[s1[i]]++;
+    //Delete one count from the ascii position of the current character of the second string
+    counts[s2[i]]--;
+
+  }
+
+  //Check if any value in the vector is not 0
+
+  for (int i = 0; i<counts.size(); ++i){
+
+    if (counts[i]!=0){
+      return false;
+
+    }
+  }
+  return true;
+}
+bool anagrams_caseInsensitive (string & s1, string & s2){
+
+  //Create the equivalent low caps strings for each string
+  string lowCaps_s1 = "";
+  string lowCaps_s2 = "";
+
+//for each character in string 1
+  for (int i=0; i<s1.size(); ++i){
+    //if it is a mayus letter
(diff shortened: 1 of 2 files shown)


laiaaa0/assignment1/q1.cpp
```diff
@@ -0,0 +1,77 @@
...
+  else{
+    if (anagrams_caseInsensitive(s1,s2)){
+      cout<<"These strings are case-insensitive-anagrams!"<<endl;
+
+    }
+    else {
+      cout<<"These strings are not anagrams"<<endl;
+    }
+  }
+}
```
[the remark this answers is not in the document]
laiaaa0: I believe it still works with sentences, although they must have the same number of spaces and punctuation marks for them to be anagrams. The method will still be of counting the ascii values so it still works? I haven't found a counter example yet.

laiaaa0/assignment1/q1.cpp
```diff
@@ -1,13 +1,25 @@
...
+  cost : O (n)
+  )
+*/
+bool anagrams_caseSensitive (const string & s1, const string & s2){
   // Compare lengths. If they are not the same, return false.
   if (s1.size() != s2.size()) return false;
 
-  // Create a vector of the size of the ascii table.
-  std::vector<int> counts(128);
+  // Create a vector of the size of the EXTENDED ascii table.
```
ssancho2: I don't want to insist much on this, just remember that it's "a bit" more complex than that. :)
Did you try running it and entering non-ascii characters?

laiaaa0/assignment1/q2.cpp
```diff
@@ -1,126 +1,127 @@
...
-      return next;
+    ListNode * GetNext (){
+      return this->node_next_;
     }
-    void setNext (ListNode *n){
-      this->next = n;
+    void SetNext (ListNode *n){
+      this->node_next_ = n;
+      /*
+TODO
```
ssancho2: The safest way to do this is to make the node's next field private (and not provide a setter). Then mark the list class as a friend of node so it's the only one that can actually access and modify it.
Then of course you need to provide methods to create and modify lists.

laiaaa0/assignment1/q2.cpp
```diff
@@ -1,126 +1,127 @@
...
-    T value;
-    ListNode * next;
+    T node_value_;
+    ListNode * node_next_;
   public:
     ListNode(){
-
-        this->next = NULL;
-
+        this->node_next_ = NULL;
```
ssancho2: Forgot to mention before: nullptr is usually preferred to NULL, since it's typed. (and NULL == 0)

laiaaa0/assignment1/q2.cpp
```diff
@@ -1,126 +1,127 @@
...
-    newNode->setValue(i);
-    current->setNext(newNode);
+    newNode->SetValue(i);
+    current->SetNext(newNode);
     current = newNode;
   }
   MyList<int> l (initial);
-  cout<<"The list is:"<<endl;
-  l.printList();
+  //cout<<"The list is:"<<endl;
```
ssancho2: Uncomment or delete. :)

PR #4, opened 2018-05-20
Add files via upload

marahgh:


marahgh/assignment1/Q1.cpp
```diff
@@ -0,0 +1,129 @@
...
+	able to handle both case sensitive and case insensitive anagrams */
+bool areAnagrams(string s1, string s2){
+    if(s1.size()!= s2.size()){
+        return false;
+    }
+    int histogram[LEN];
+    for(int i=0;i<LEN;i++){
+        histogram[i]=0;
+    }
+    for(int i=0;i<s1.size();i++){
```
ssancho2: it may be more readable to use the new style iterators: `for (char c : s1) { ... } `

marahgh/assignment1/Q2.cpp
```diff
@@ -0,0 +1,128 @@
...
+        if(curr->getNext()==nullptr){
+            delete head;
+            head=nullptr;
+            tail=nullptr;
+            size=0;
+            return;
+        }
+        while (curr){
+            if(curr->getNext()== tail){
+                beforeLast=curr;
```
ssancho2: NIT: You don't need the extra variable, you can use curr directly after the loop.

PR #5, opened 2018-05-21
MennaMohie-1

MennaMohie:


MennaMohie/assignment1/Q1.cpp
```diff
@@ -0,0 +1,134 @@
+#include<bits/stdc++.h>
+#include<stdlib.h>
+#define endl '\n'
```
ssancho2: Why ?
The standard library already defines a endl

MennaMohie: It's because <<'\n' is faster than <<endl but endl is easier to type so I defined it as '\n' because it used to make a difference when I solved problems on CodeForces, but I will remove it. :D

MennaMohie/assignment1/Q1.cpp
```diff
@@ -0,0 +1,134 @@
...
+        if(sensitivity=='i')
+        {
+            word1[i]=tolower(word1[i]);
+            word2[i]=tolower(word2[i]);
+        }
+        word1_map[word1[i]].push_back(i);
+        word2_map[word2[i]].push_back(i);
+    }
+
+    bool differentOrder=false;
```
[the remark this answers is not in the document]
MennaMohie: I think I should modify the strings to be able to use (word1!=word2) for the case insensitive checking?

MennaMohie:
@ssancho2 I've made the requested changes. :D

MennaMohie/assignment1/Q1.cpp
```diff
@@ -23,7 +22,15 @@ int main()
         cout<<"Please type 's' for Case Sensitive Anagram Checking or 'i' for Case Insensitive Anagram Checking: ";
         cin>>sensitivity;
 
-        cout<<(anagramWordCheck(word1, word2, tolower(sensitivity))?"The two words are anagrams.":"The two words are not anagrams.");
+        bool caseSensitive;
+        sensitivity=tolower(sensitivity);
+        if(sensitivity=='i')
+            caseSensitive=false;
+        else if (sensitivity=='s')
+            caseSensitive=true;
```
ssancho2: else?
I think lines 25-30 can be more clear as:
`bool caseSensitive = tolower(sensitivity) == 's'`

MennaMohie/assignment1/Q1.cpp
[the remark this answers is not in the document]
ssancho2: Yes, you are right.

MennaMohie/assignment1/Q2.cpp
```diff
@@ -27,37 +26,88 @@ class SinglyLinkedList
     int Size;
 
 public:
-    SinglyLinkedList():head(NULL),tail(NULL),Size(0) {}
+    SinglyLinkedList():head(nullptr),tail(nullptr),Size(0) {}
+    SinglyLinkedList(const SinglyLinkedList<T> &SLL)
+    {
+        this->Size = SLL.Size;
```
ssancho2: Why not `Size = SLL.Size;` ?

MennaMohie/assignment1/Q2.cpp
```diff
@@ -27,37 +26,88 @@ class SinglyLinkedList
...
     }
 
-    void DisplayAt(int pos)
+    T GetKthToLastNode(int k)
     {
-        assert(pos>=0 && pos<Size);
-        int i=0;
-        Node<T> *temp = head;
-        while(temp!=NULL)
+        if(k>=0 && k<Size);
```
ssancho2: That extra `;` is playing tricks on you :)

MennaMohie/assignment1/Q2.cpp
```diff
@@ -27,37 +26,88 @@ class SinglyLinkedList
...
+                other_list_ptr=other_list_ptr->next;
+                this_temp_ptr=this_temp_ptr->next;
+            }
+            this->tail = this_temp_ptr;
+        }
+        return *this;
     }
 
-    void DisplayAt(int pos)
+    T GetKthToLastNode(int k)
```
ssancho2: This declaration leaves you no room to return a error in case k is out of bounds, either:
* return a success bool, and have a `T *` parameter to put the node. (preferred)
* return a `const T *`, using nullptr as error/out of bounds.
* Throw an exception (not recommneded)

PR #6, opened 2018-05-22
JoodHajeer-1

JoodHajeer:


JoodHajeer/assignment1/Q1
```diff
@@ -0,0 +1,69 @@
+#include <iostream>
+#include <stdio.h>  
+#include <vector>
+#include <string>
+#include <map>
+#include <algorithm>
+using namespace std;
+map<string,int> mp; //maps a string to its frequency in the sentence
```
ssancho2: Global variables should be avoided if possible.
Could use a better name too :)

JoodHajeer/assignment1/Q1
```diff
@@ -0,0 +1,69 @@
...
+void toLower(string &s){ //change all capital letters in s to lower letters
+	for(int i=0;i<s.size();i++)
+		if(s[i]>='A'&&s[i]<='Z')
```
ssancho2: You can use `tolower` from `<cctype>`

JoodHajeer/assignment1/Q1
```diff
@@ -0,0 +1,69 @@
...
+}
+bool areAnagrams(string s,string t, bool caseSensitive){ 
+	//checks if two sentences are anagrams 
+	mp.clear();
+	if(!caseSensitive){
+		toLower(s);
+		toLower(t);
+	}
+	bool f=seperateWords(s,1);
+	f=seperateWords(t,-1);
```
ssancho2: You also need to check that all counts in `mp` are 0 now, right?

JoodHajeer/assignment1/Q1
```diff
@@ -0,0 +1,69 @@
...
+	return f;
+}
+int main() {
+	#ifndef ONLINE_JUDGE
```
ssancho2: :)

JoodHajeer/assignment1/Q1
```diff
@@ -0,0 +1,69 @@
+#include <iostream>
+#include <stdio.h>  
+#include <vector>
+#include <string>
+#include <map>
+#include <algorithm>
+using namespace std;
+map<string,int> mp; //maps a string to its frequency in the sentence
+void toLower(string &s){ //change all capital letters in s to lower letters
+	for(int i=0;i<s.size();i++)
```
ssancho2: You can use `for (char & c : s) { ... } `

JoodHajeer/assignment1/Q2
```diff
@@ -0,0 +1,113 @@
...
+class SLLNode{
+//Singly Linked List Node Class building block for a list
+// A single linked list node class to store elements of a generic type T
+public:
+SLLNode(T &VAL, SLLNode* NEXT);
+~SLLNode();
+private:
+	T val;
+	SLLNode * next;
+	friend class SLList<T>;
```
ssancho2: Nice :)

JoodHajeer/assignment1/Q2
```diff
@@ -0,0 +1,113 @@
...
+	for(p=head;p!=NULL;p=p->next)//empty list will not enter this loop
+		count++;
+	return count;
+}
+template <class T>
+T SLList<T> :: findElementToLast(int k){//assuming k belongs to the interval [0,size)
+	int sz=this->size();//gets the current size of the list
+	k=sz-k;//kth element to last is (size-k)th element to first
+	if(k<=0){
+		return -1;//handles cases when k has a wrong value or list is empty
```
ssancho2: This limits the possible types T can be, right?

JoodHajeer/assignment1/Q2
```diff
@@ -0,0 +1,113 @@
...
+template <class T>
+int SLList<T> :: size(){//computes and return the current size of the list
+	int count=0;
+	SLLNode<T> * p;
+	for(p=head;p!=NULL;p=p->next)//empty list will not enter this loop
+		count++;
+	return count;
+}
+template <class T>
+T SLList<T> :: findElementToLast(int k){//assuming k belongs to the interval [0,size)
```
ssancho2: It will be nice if the code actually checks those limits.

JoodHajeer/assignment1/Q2
```diff
@@ -0,0 +1,113 @@
...
+	#ifndef ONLINE_JUDGE
+	freopen("input.txt", "r", stdin);
+	#endif
+	//a quick test 
+	SLList<int> list;
+	for(int i=0;i<=10;i++)
+		list.addToTail(i);
+	for(int i=0;i<=10;i++){
+		cout<<list.findElementToLast(i)<<" ";
+	}
```
ssancho2: What will happen if you do `SLList<int> list2 = list;` ?

JoodHajeer/assignment1/Q2
```diff
@@ -0,0 +1,113 @@
...
+	void clear();
+	int size();
+	T findElementToLast(int k);
+private:
+	SLLNode<T> * head;
+	SLLNode<T> * tail;
+};
+template <class T>
+SLList<T> :: SLList(){
+	head=tail=NULL;
```
ssancho2: It's generally preferred to use nullptr instead of NULL, it's typed when NULL is actually 0.

JoodHajeer/assignment1/Q1
```diff
@@ -4,66 +4,68 @@
...
 #include <algorithm>
+#include <cctype>
 using namespace std;
-map<string,int> mp; //maps a string to its frequency in the sentence
-void toLower(string &s){ //change all capital letters in s to lower letters
-	for(int i=0;i<s.size();i++)
-		if(s[i]>='A'&&s[i]<='Z')
-			s[i]+='a'-'A';
+
+void toLower(string &s) { //change all capital letters in s to lower letters
```
ssancho2: Minor comment, it's a better to use pointers instead of references for parameters you want to modify. It's a bit more cumbersome to write all the derreferences, but it makes crystal clear for the caller that the variable will be modified, compare:
toLower(s) vs toLower(&s)

JoodHajeer/assignment1/Q1
```diff
@@ -4,66 +4,68 @@
...
 }
-bool seperateWords(string s,int inc){ 
-//builds the words frequency map
-//for the first call the functoin adds 1 to the word frequency
-//for the second call the function subtracts 1 from the word frequency
+bool seperateWords(string s, map<string, int> & wordsFrequency, int inc) {
+	//builds the words frequency map
+	//for the first call the functoin adds 1 to the word frequency
+	//for the second call the function subtracts 1 from the word frequency
 	string tmp;
```
ssancho2: Maybe 'word' instead of 'tmp' ? :)

JoodHajeer/assignment1/Q2
```diff
@@ -42,72 +44,84 @@ private:
...
-	if(k<=0){
-		return -1;//handles cases when k has a wrong value or list is empty
-	}
-	SLLNode <T> * p=head;
-	for(int i=1; i<k ;i++)//searches for elemnt at position k to last
-		p=p->next;
+T SLList<T> ::findElementToLast(int k) {
+	int sz = this->size();//gets the current size of the list
+	if (k < 0 || k >= sz) //handles cases when k has a wrong value or list is empty
+		exit(1);
```
ssancho2: A bit extreme for not finding a element in a list :)
Options are either return a const pointer (and nullptr for element not found), or return a bool to indicate success and have a out parameter for the node, a declaration like:
bool findElementToLast(int k, SLLNode *element)

JoodHajeer/assignment1/Q2
ssancho2: Right. The two list will have the same pointer, and when their destructors are called, the first will be fine and the second will crash freeing already freed memory.
Indeed this is a shallow copy, and what's needed is a deep copy (making a copy of each node and the list object).
Another option is to declare the copy constructor and assignment as `= delete`, so the compiler will throw a error when somebody tries to make a copy with `=`.
Note that a similar problem will happen with the move constructor/assignment.

PR #7, opened 2018-05-22
Added problem 1 and problem 2 answers

MayssaraOmar:


MayssaraOmar/assignment1/Problem 2.cpp
```diff
@@ -0,0 +1,178 @@
...
+        T data;
+        Node *next;
+        Node(const T &input, Node *next) : data(input), next(next) {}
+    };
+    int listsize;
+    Node *head;
+    Node *tail;
+
+
+    struct empty_list_exception : public exception {
```
ssancho2: Just a side note to be careful with exceptions. They are a powerful tool, but as most tools in the c++ toolkit it's easy to do more harm than good :)

MayssaraOmar/assignment1/Problem 1.cpp
```diff
@@ -0,0 +1,71 @@
...
+        for(int j=0; j<words[i].length(); j++){
+            letters_freq[toupper(words[i][j])]++; // toupper Converts the char to its uppercase equivalent if c is a lowercase letter, else the char value returned is unchanged.
+        }
+        words_freq[letters_freq]++;
+    }
+    return   words_freq;
+
+}
+
+//returns 0 if both sentences are not anagrams
```
ssancho2: Better to use false and true instead of 0 and 1 here.

MayssaraOmar/assignment1/Problem 2.cpp
```diff
@@ -0,0 +1,178 @@
...
+    struct out_of_bounds_exception : public exception {
+        const char * what () const throw() {
+            return "index out of bounds";
+        }
+    };
+
+
+public:
+    linkedlist() : head(NULL), tail(NULL), listsize(0) {} //first constructor
+    linkedlist(Node *newnode) : head(newnode), tail(newnode), listsize(1) {} //second constructor
```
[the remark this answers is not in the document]
MayssaraOmar: Can you please clarify this part "note that the compiler will automatically convert Node * s into linked lists when needed."?

MayssaraOmar/assignment1/Problem 2.cpp
MayssaraOmar: It's the first time I use exceptions in C++. Is this a good exceptions implementation ? And what is the best way to handle these kind of errors instead of using exceptions?

MayssaraOmar/assignment1/Problem 1.cpp
```diff
@@ -18,54 +18,71 @@ vector<string> split_spaces(string sentence){
...
     vector<string> words = split_spaces(sentence1);
     map< vector<int>,int > words_freq;
-    for(int i=0; i<words.size(); i++){
-        vector<int> letters_freq(128); //The ASCII table has 128 characters
-        for(int j=0; j<words[i].length(); j++){
-            letters_freq[toupper(words[i][j])]++; // toupper Converts the char to its uppercase equivalent if c is a lowercase letter, else the char value returned is unchanged.
+    for (const string &word : words) {
+        vector<int> chars_freq(128); //The ASCII table has 128 characters
+        for (char c : word){
+            if(c<0 || c>127){
```
ssancho2: Nice. That is one way to don't care if char is signed or unsigned :)

MayssaraOmar/assignment1/Problem 1.cpp
```diff
@@ -18,54 +18,71 @@ vector<string> split_spaces(string sentence){
...
         last = next+1;
     }
 
     return tokens;
 
 }
 
-map< vector<int>,int > get_map(string sentence1){
-
+map< vector<int>,int > get_map(string sentence1, bool *error){
```
ssancho2: Usually, it's easier to return bool for success and use the out parameter for the map in this case. That way you can call the function on a if statement to handle the error.

MayssaraOmar/assignment1/Problem 1.cpp
```diff
@@ -18,54 +18,71 @@ vector<string> split_spaces(string sentence){
...
 
 int main()
 {
     // test
-    cout << are_anagrams( "  Good   morning   world " , "dooG gninrom dlrow") << endl;
-    cout << are_anagrams( "GoOd" , "dogo") << endl;
-    cout << are_anagrams( "GoO,d" , "dogo,") << endl;
-    cout << are_anagrams( "GoOd," , "dogo ,") << endl;
-    cout << are_anagrams( " GoOd ," , " , dogo") << endl;
+    bool error;
```
ssancho2: Maybe you should check error... :) 
Or pass nullptr if you don't care...

MayssaraOmar/assignment1/Problem 1.cpp
```diff
@@ -18,54 +18,71 @@ vector<string> split_spaces(string sentence){
...
+bool are_anagrams(string sentence1, string sentence2, bool *error){
+    if(error){
+      *error = 0;
+    }
 
-    map< vector<int>,int > map1 = get_map(sentence1);
-    map< vector<int>,int > map2 = get_map(sentence2);
+    map< vector<int>,int > map1 = get_map(sentence1, error);
+    if(error && *error){
+        return 0;
```
ssancho2: `return false`?

MayssaraOmar/assignment1/Problem 2.cpp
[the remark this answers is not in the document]
ssancho2: One of the many features of c++ to shot you in the foot :)
A constructor with only one parameter is interpreted by the compiler as a conversion function.
For the long version, check the 'explicit' keyword: https://en.cppreference.com/w/cpp/language/explicit

MayssaraOmar/assignment1/Problem 2.cpp
ssancho2: I'm not sure, I haven't used them at all here :)
Looks reasonable.
Usually, we use a bool return indicating success and use out parameters for the real return values. Another option is when returning pointers you can use nullptr to indicate a failure.
When things get more complicated, you can return a 'result' type, that carries information about WHAT went wrong in case of error.

PR #8, opened 2018-05-24
Lacrit 1

Lacrit:


Lacrit/assignment1/q1.cpp
```diff
@@ -0,0 +1,49 @@
+#include <iostream>
+#include <unordered_map>
+#include <boost/algorithm/string.hpp>
+
+
+// input : string a , string b || output : bool
```
ssancho2: Please move the comment to the function it applies to.

Lacrit/assignment1/q1.cpp
```diff
@@ -0,0 +1,49 @@
+#include <iostream>
+#include <unordered_map>
+#include <boost/algorithm/string.hpp>
+
+
+// input : string a , string b || output : bool 
+// determine if one is an anagram of another 
+	
+std::unordered_map<char, int> count_occurences(std::string& a, const bool c)
```
ssancho2: Parameters can use better names :)
Specially bools, that could be one thing or the exact opposite.

Lacrit/assignment1/q1.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+	//case insensitive
+	if (!c)
+		boost::algorithm::to_lower(a);	
+		
+	//to store letter frequency in the first string
+	std::unordered_map<char,int> count;
+
+	//count char occurences
+	for (size_t i = 0; i < a.length(); i++)
```
ssancho2: You could do: `for (char c : a) { ... }` (if you rename the bool parameter, of course)

Lacrit/assignment1/q1.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+int main() 
+{
+	std::string a, b;
+	bool c; 
+	std::getline(std::cin, a);
+	std::getline(std::cin, b);
+	remove_spaces(a);
+	remove_spaces(b);
+	std::cin >> c;
```
ssancho2: NIT: You could print a line explaining to the user to enter two lines and then if he/she wants case sensitive or case insensitive anagram checking. For the last one it'll be better to use a char input, like 'c' for case sensitive and 'i' for case insensitive.

Lacrit/assignment1/q1.cpp
```diff
@@ -0,0 +1,49 @@
...
+	std::cout << isAnagram(a, b, c) << std::endl;
```
ssancho2: NIT: Similarly for the output here, a textual output like "They are anagrams"/ "They aren't anagrams". Or even "Yes" / "No" :)

Lacrit/assignment1/q2.cpp
```diff
@@ -0,0 +1,60 @@
+#include <iostream>
+
+
+template<class T>
```
ssancho2: A note on interface design, letting the user modify next freely it's risky.

Lacrit/assignment1/q2.cpp
```diff
@@ -0,0 +1,60 @@
...
+  Node<char>* head = NULL;
+
+  // create linked r->n->b->a
+  push(&head, 'a');
+  push(&head, 'b');
+  push(&head, 'n');
+  push(&head, 'r');
+
+  std::cout << findNode(head, 3) << std::endl;
+  return 0;
```
ssancho2: You need to free all nodes.

Lacrit/assignment1/q2.cpp
```diff
@@ -0,0 +1,60 @@
...
+struct Node
+{
+    T i;
+    struct Node *next;
+};
+
+
+template<class T> 
+T findNode (Node<T> *listHead, int n)
+{
```
ssancho2: What happens if n is out of range?
What if the list is empty?

Lacrit/assignment1/q2.cpp
```diff
@@ -0,0 +1,60 @@
...
+            ptr2 = ptr2->next;
+        }
+   }
+   return ptr2->i;
+}
+
+template<class T> 
+void push(Node<T>** head, T data)
+{
+  Node<T>* new_node =(Node<T>*) malloc(sizeof(Node<T>));
```
ssancho2: Why `malloc` and not `new`?

Lacrit/assignment1/q1.cpp
```diff
@@ -37,13 +33,22 @@ void remove_spaces(std::string& str)
...
 	std::getline(std::cin, a);
+	std::cout << "Input second string: " << std::endl;
 	std::getline(std::cin, b);
-	remove_spaces(a);
-	remove_spaces(b);
-	std::cin >> c;
-	std::cout << isAnagram(a, b, c) << std::endl;
+	std::cout << "Case Sensitive?(c|i) (c => sensitive || i => insensitive)" << std::endl;
+	std::cin >> case_sensitive;
+	const std::string word = "Words aren`t anagrams";
```
ssancho2: (*) word is a weird name for the message.

Lacrit/assignment1/q1.cpp
```diff
@@ -37,13 +33,22 @@ void remove_spaces(std::string& str)
...
+		remove_spaces(a);
```
ssancho2: (*) Please don't change the indentation level in the same block.

Lacrit/assignment1/q1.cpp
```diff
@@ -37,13 +33,22 @@ void remove_spaces(std::string& str)
...
+		remove_spaces(b);
```
ssancho2: I think both remove_space calls and the check if they are the same length should be part of `isAnagram`

Lacrit/assignment1/q1.cpp
```diff
@@ -37,13 +33,22 @@ void remove_spaces(std::string& str)
...
-	std::cin >> c;
-	std::cout << isAnagram(a, b, c) << std::endl;
+	std::cout << "Case Sensitive?(c|i) (c => sensitive || i => insensitive)" << std::endl;
+	std::cin >> case_sensitive;
+	const std::string word = "Words aren`t anagrams";
+		remove_spaces(a);
+		remove_spaces(b);	
+		if (a.length() == b.length())
+		{
+	   	    bool res = isAnagram(a, b, case_sensitive);
```
ssancho2: (*) case_sensitive is a char ('c' or 'i'), the function expects a bool.

Lacrit/assignment1/q2.cpp
```diff
@@ -4,44 +4,76 @@
...
+	public: 
+		Node* get_next()
+		{
+			return next;
+		}  
+		T get_i()
+		{
+			return i;
+		}
+		void set_next(Node *new_next)
```
ssancho2: The user can still modify next :)
One solution is to make a LinkedList class, and mark is as a friend of Node, that way only there you can modify it.

Lacrit/assignment1/q2.cpp
```diff
@@ -4,44 +4,76 @@
...
+			i = j;
+		}
+			
 };
 
 
 template<class T> 
 T findNode (Node<T> *listHead, int n)
 {
+	 if (listHead == NULL) return 0;
```
ssancho2: nullptr is usually preferred to NULL, since nullptr is typed, while NULL is a int with value 0
Also, returning 0 severely limits the types T can be :)

Lacrit/assignment1/q2.cpp
```diff
@@ -4,44 +4,76 @@
...
 template<class T> 
 void push(Node<T>** head, T data)
 {
-  Node<T>* new_node =(Node<T>*) malloc(sizeof(Node<T>));
- 
-  new_node->i  = data;
-  new_node->next = (*head);
+  //Node<T>* new_node =(Node<T>*) malloc(sizeof(Node<T>))
+  //Just left the upper row for reference. Correction:
+  Node<T>* new_node = new Node<T>(); // no need to pass the size
```
ssancho2: If you add a constructor to Node, you can replace this 3 lines with:
`Node<T>* new_node = new Node<T>(data, *head);`
Or even the whole function with:
`head = new Node<T>(data, *head);`
:)

Lacrit/assignment1/q2.cpp
```diff
@@ -4,44 +4,76 @@
...
 
 template<class T> 
 T findNode (Node<T> *listHead, int n)
 {
+	 if (listHead == NULL) return 0; 
      Node<T> *ptr1, *ptr2;  // we need 2 pointers
      ptr1 = ptr2 = listHead; // set the pointers to point to the list head initially
 
-    while(ptr1->next != NULL) // keep looping until we reach the tail (next will be NULL for the last node)
+    while(ptr1->get_next() != NULL) // keep looping until we reach the tail (next will be NULL for the last node)
```
ssancho2: What happens when n is less than 0 or bigger (or equal) to the list size?

PR #9, opened 2018-05-26
Marah assignment1

marahgh:


marahgh/assignment1/Q1.cpp
```diff
@@ -0,0 +1,171 @@
...
+            histogram[lowerLetter - 'a']++;
+            continue;
+        }
+        if(c <='a' && c>='z'){
+            histogram[c - 'a']++;
+        }else{
+           histogram[c - 'A'+ LEN]++; 
+        }
+    }
+    for(char c: s2){
```
ssancho2: Lines 37-52 are very similar to 22-36. Maybe this can be extracted into a function?
It also looks like the code could be a bit simpler, maybe?

marahgh/assignment1/Q1.cpp
```diff
@@ -0,0 +1,171 @@
...
+            count++;
+        }
+    }
+    return count+1;
+}
+
+string getKWord(string sentence, int k){
+    int start, end,count=0;
+    int wordsNum=countWords(sentence);
+    if(k<=0 || sentence.empty() || sentence== ""){
```
ssancho2: `sentence.empty()` and `sentence== ""` are equivalent.

marahgh/assignment1/Q1.cpp
```diff
@@ -0,0 +1,171 @@
...
+        }
+        i++;
+    }
+    return sentence.substr(start,end-start+1);
+}
+
+
+vector<string> splitIntoWords(string sentence){
+    vector<string> words=vector<string>();
+    for(int i=0;i<=countWords(sentence);i++){
```
ssancho2: I think it should start at i=1, since getKWord returns the empty string for k=0.

marahgh/assignment1/Q1.cpp
```diff
@@ -0,0 +1,171 @@
...
+    
+    for(int i=0;i<sentence2.size();i++){
+        for(string word1: sentence1){
+            if(areAnagrams(sentence2[i],word1,caseSensetive)){
+                counts2[i]++;
+            }
+        }
+    }
+    
+    /*checking for every sentence if every word
```
ssancho2: This doesn't quite work, for example 'art rat' and 'tar atr' should be considered anagrams, right?

marahgh/assignment1/Q1.cpp
```diff
@@ -0,0 +1,171 @@
...
+        i++;
+    }
+    return sentence.substr(start,end-start+1);
+}
+
+
+vector<string> splitIntoWords(string sentence){
+    vector<string> words=vector<string>();
+    for(int i=0;i<=countWords(sentence);i++){
+        words.push_back(getKWord(sentence,i));
```
ssancho2: Note that this is sub-optimal, since for each call getKWord will go through the string until it finds the requested word. This is O(n^2), when it could be O(N).

marahgh/assignment1/Q2.cpp
```diff
@@ -0,0 +1,107 @@
...
+	private:
+	Node<T>* head;
+	Node<T>* tail;
+	int size;
+	
+	public:
+	
+	List():head(nullptr),tail(nullptr),size(0){}
+	
+	Node<T>* getHead(){
```
ssancho2: Returning a Node* lets the user do whatever it wants with it. It's safer if the interface of the list just receives/returns objects of type T

marahgh/assignment1/Q2.cpp
```diff
@@ -0,0 +1,107 @@
...
+        while (currentNode){
+            Node<T>* nextNode = currentNode->next;    
+            delete currentNode;                         
+            currentNode = nextNode;    
+        }
+    }
+//----------------------------------------------------------------------------------//
+/* A Method to return the kth to last element of a singly linked list*/
+
+T* getKthToLast(int k){
```
ssancho2: Very concise! Nice.

marahgh/assignment1/Q1.cpp
```diff
@@ -0,0 +1,171 @@
...
+/*A function that determines if two strings are anagrams ,
+	able to handle both case sensitive and case insensitive anagrams,
+	according to value of caseSensetive flag variable */
+bool areAnagrams(string s1, string s2, bool caseSensetive){
+    const int LEN= 'z' -'a' +1;
+
+    if(s1.size()!= s2.size()){
+        return false;
+    }
+    vector<int> histogram(LEN*2,0);
```
ssancho2: Note that the code will be much simpler if you actually don't have the restriction on only english letters :)

marahgh/assignment1/Q1.cpp
```diff
@@ -8,49 +7,54 @@ using namespace std;
...
-	able to handle both case sensitive and case insensitive anagrams,
-	according to value of caseSensetive flag variable */
-bool areAnagrams(string s1, string s2, bool caseSensetive){
-    const int LEN= 'z' -'a' +1;
-
-    if(s1.size()!= s2.size()){
-        return false;
+/*auxiliary function for updating the given histogram, according to the value of the flag add
+  increments the counters if its true and decrements them otherwise*/
+bool updateHistogram(vector<int>& histogram,string word,bool add, bool caseSensetive){
```
ssancho2: A note on interface design: Having a bool parameter can be confusing (is easy to think one way or the other, like it was true for add or true for substract). Having two in a row is double confusing :)

marahgh/assignment1/Q1.cpp
```diff
@@ -8,49 +7,54 @@ using namespace std;
...
+     const int LEN= 'z' -'a' +1;
+    int factor=0;
```
ssancho2: The parameter could be factor directly, instead of add.

PR #11, opened 2018-06-04
MennaMohie-2

MennaMohie:


MennaMohie/assignment2/Solution.cpp
```diff
@@ -0,0 +1,174 @@
...
+        return root;
+    }
+
+    bool contains(Node<T> *currentRoot, T val)
+    {
+        if(currentRoot == nullptr)
+            return false;
+        if(currentRoot->value == val)
+            return true;
+        return (contains(currentRoot->left, val) || contains(currentRoot->right, val));
```
ssancho2: NIT: Extra unneeded parenthesis.

MennaMohie/assignment2/Solution.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+template<class T>
+class Node
+{
+    T value;
+    Node *left, *right;
+    friend class BinaryTree<T>;
+
+public:
+    Node():left(nullptr),right(nullptr) {}
```
ssancho2: Do you need this constructor?

MennaMohie/assignment2/Solution.cpp
```diff
@@ -0,0 +1,174 @@
...
+    bool contains(Node<T> *currentRoot, T val)
+    {
+        if(currentRoot == nullptr)
+            return false;
+        if(currentRoot->value == val)
+            return true;
+        return (contains(currentRoot->left, val) || contains(currentRoot->right, val));
+    }
+
+    bool resetRoot(T val)
```
ssancho2: Why this function?

MennaMohie/assignment2/Solution.cpp
```diff
@@ -0,0 +1,174 @@
...
+            return true;
+        }
+        else
+            return false;
+    }
+
+    //adds a new node (val) as a left child to an existing node (parentVal)
+    bool addAsLeftChild(Node<T> *currentRoot, T parentVal, T val)
+    {
+        if(!contains(root, parentVal) || contains(root, val) || currentRoot==nullptr)
```
ssancho2: Note that the function `contains` explores the whole tree, and you will do this on every step of the way.
It's better to split this function in 2, a public one that does the sanity checking once (and doesn't need the currentRoot parameter), and a private function that actually traverses the tree and does the insertion.
The same applies to the function leastCommonAncestor bellow.

MennaMohie/assignment2/Solution.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+public:
+    BinaryTree(T val):root(new Node<T>(val)) {}
+
+    Node<T>* getRoot()
+    {
+        return root;
+    }
+
+    bool contains(Node<T> *currentRoot, T val)
```
ssancho2: Note that it may be more useful to return the node found if it exists, and nullptr otherwise.

MennaMohie/assignment2/Solution.cpp
```diff
@@ -0,0 +1,174 @@
...
+                return false;
+            }
+        }
+        if(addAsLeftChild(currentRoot->left, parentVal, val))
+            return true;
+        return addAsLeftChild(currentRoot->right, parentVal, val);
+
+    }
+
+    bool addAsRightChild(Node<T> *currentRoot, T parentVal, T val)
```
ssancho2: If possible, it will be nice to merge the addAsLeftChild & addAsRightChild functions into one (with a extra parameter to differentiate). This extra parameter can be a bool, or a enum (preferred).

MennaMohie/assignment2/Solution.cpp
```diff
@@ -0,0 +1,174 @@
...
+            return false;
+
+        if(currentRoot->value==val)
+            return true;
+
+        bool foundInLeft = printAncestors(currentRoot->left, val);
+        bool foundInRight = printAncestors(currentRoot->right, val);
+        if(foundInLeft || foundInRight)
+        {
+            cout<<currentRoot->value<<endl;
```
ssancho2: Please return the path from the printAncestors function, and print it on main. This way your function can be reused.
Since you are already using the return for success/failure, the easies way to do this is to have a `vector<T> *` out parameter.

MennaMohie/assignment2/Solution.cpp
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+        if(addAsRightChild(currentRoot->left, parentVal, val))
+            return true;
+        else
+            return addAsRightChild(currentRoot->right, parentVal, val);
+
+    }
+
+    void traverse(Node<T> *currentRoot)
```
ssancho2: NIT: The tree will be more visible if you add identation according to the level, changing the function to:
`void traverse(Node<T> *currentRoot, string identation = "")`
add identation to the output, before the value, and passing identation + "  " to the recursive calls.
You may also add output placeholder (maybe '*') where "null nodes" are. (to differentiate a trees with only one child on one side or the other)

MennaMohie/assignment2/Solution.cpp
```diff
@@ -152,22 +147,40 @@ class BinaryTree
...
-    tree.addAsLeftChild(tree.getRoot(), 2, 4);
-    tree.addAsRightChild(tree.getRoot(), 2, 5);
-    tree.addAsLeftChild(tree.getRoot(), 3, 6);
-    tree.addAsRightChild(tree.getRoot(), 3, 7);
-    tree.addAsLeftChild(tree.getRoot(), 7, 8);
-    tree.addAsRightChild(tree.getRoot(), 8, 9);
-
-    tree.printAncestors(tree.getRoot(), 9);
-    cout<<endl;
+    tree.addChildToParent(tree.getRoot(), 1, 2, addLeft);
```
ssancho2: It's kind of a strange interface when you are calling a method on the tree and still need to pass the root to it. It kind of knows it's own root :)

MennaMohie/assignment2/Solution.cpp
```diff
@@ -12,118 +14,111 @@ class Node
...
-            {
-                return false;
-            }
-        }
-        if(addAsRightChild(currentRoot->left, parentVal, val))
-            return true;
         else
-            return addAsRightChild(currentRoot->right, parentVal, val);
-
+            add(currentRoot, parentVal, val, dir);
```
ssancho2: Missing `return true`?

MennaMohie/assignment2/Solution.cpp
```diff
@@ -12,118 +14,111 @@ class Node
...
+    }
+public:
+    BinaryTree(T val):root(new Node<T>(val)),Size(1) {}
+
+    Node<T>* getRoot()
+    {
+        return root;
+    }
 
+    Node<T>* contains(Node<T> *currentRoot, T val)
```
ssancho2: NIT: function name suggest just a bool return indicating if the node is in the tree.
For a function actually returning the node something like findNode, getNode, etc. will be better.

MennaMohie/assignment2/Solution.cpp
```diff
@@ -12,118 +14,111 @@ class Node
...
-
-    //adds a new node (val) as a left child to an existing node (parentVal)
-    bool addAsLeftChild(Node<T> *currentRoot, T parentVal, T val)
-    {
-        if(!contains(root, parentVal) || contains(root, val) || currentRoot==nullptr)
-            return false;
         if(currentRoot->value==parentVal)
         {
-            if(currentRoot->left==nullptr)
+            if(dir==0)
```
ssancho2: Since you have the enum, better to use it :)
`dir == addLeft` looks better, right?

MennaMohie/assignment2/Solution.cpp
```diff
@@ -12,118 +14,111 @@ class Node
...
+        if(currentRoot==nullptr || currentRoot->value==val)
+            return currentRoot;
+        Node<T> *leftSide = contains(currentRoot->left, val);
+        Node<T> *rightSide = contains(currentRoot->right, val);
+        return (leftSide==nullptr?rightSide:leftSide);
     }
 
-    bool addAsRightChild(Node<T> *currentRoot, T parentVal, T val)
+    //adds a new node (val) as a left/right child (depends on dir) to an existing node (parentVal)
+    bool addChildToParent(Node<T> *currentRoot, T parentVal, T val, direction dir)
```
ssancho2: You don't need the currentRoot parameter, just pass root to the add method bellow.

PR #12, opened 2018-06-05
JoodHajeer-2

JoodHajeer:


JoodHajeer/assignment2/JoodHajeer-2.cpp
```diff
@@ -0,0 +1,252 @@
...
+	}
+	while (!ancestorsStack.empty()) {
+		T nextAncestor= ancestorsStack.top();
+		ancestorsStack.pop();
+		cout << nextAncestor << " ";
+	}
+	cout << endl;
+}
+template <class T>
+pair<bool,T> Binary_Tree<T> ::commonAncestor(T & VAL1, T & VAL2) { //complexity O(tree height)
```
ssancho2: Original way of solving it ^_^

JoodHajeer/assignment2/JoodHajeer-2.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+template <class T>
+BTNode<T> * Binary_Tree<T> ::find(T &VAL, BTNode<T> * node) {//return a pointer to the node with val if exists
+	if (!node)
+		return nullptr;
+	if (node->val == VAL)
+		return node;
+	BTNode<T> * leftSide = find(VAL, node->left);
+	BTNode<T> * rightSide = find(VAL, node->right);
+	if (leftSide)
```
ssancho2: NIT: You don't need to search the right side if already found on the left side.

JoodHajeer/assignment2/JoodHajeer-2.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+template <class T>
+BTNode<T> * Binary_Tree<T> ::getRoot() {
+	return root;
+}
+template <class T>
+Binary_Tree<T> ::Binary_Tree( Binary_Tree<T> & copy_from) {
+	BTNode<T> * currentNode = copy_from.getRoot();
+	root = nullptr;
+	if (!currentNode)
```
ssancho2: NIT: Comparing a pointer to nullptr is a bit better than casting to bool.

JoodHajeer/assignment2/JoodHajeer-2.cpp
```diff
@@ -0,0 +1,252 @@
...
+template <class T>
+class Binary_Tree{
+public:
+	Binary_Tree();
+	Binary_Tree( Binary_Tree <T> &copy_from);
+	~Binary_Tree();
+	bool insert(T &VAL);
+	bool empty();
+	void clear(); 
+	bool find(T & VAL);
```
ssancho2: Why `T&` and not `const T&` ?

JoodHajeer: It is being passed as a parameter inside the find function, can't do that if it was declared a const right?

JoodHajeer/assignment2/JoodHajeer-2.cpp
JoodHajeer: Do you prefer if I write the solution we discussed in the meeting?

JoodHajeer/assignment2/JoodHajeer-2.cpp
```diff
@@ -89,21 +85,21 @@ BTNode<T> * Binary_Tree<T> ::find(T &VAL, BTNode<T> * node) {//return a pointer
...
 		return false; 
-	if (prevNode->left != nullptr && child_ == 'L') //parent already has a left child
+	if (prevNode->left != nullptr && childDirection == false) //parent already has a left child
 		return false;
-	if (prevNode->right != nullptr && child_ == 'R')//parent already has a right child
+	if (prevNode->right != nullptr && childDirection == true)//parent already has a right child
 		return false;
 	BTNode<T> * newNode = new BTNode<T>(VAL, nullptr, nullptr, prevNode , prevNode->depth+1); //val, left, right, patent, new depth
-	if (child_=='L') 
+	if (childDirection ==false)
```
ssancho2: NIT: Comparing bools to true / false is kind of strange. Can use `if (!childDirection)` /  `if (childDirection)`
Same above.
For that to make more sense tough, the name of the bool should be indicative of what true/false means, so maybe better named as insertToTheRight, or similar.
(Ideally is better to use an enum, but don't worry :) )

PR #13, opened 2018-06-05
MayssaraOmar-2  Added assignment 2 solution

MayssaraOmar:


MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+template<class T>
+binary_tree<T>::~binary_tree() {
+	clear();
+}
+
+template<class T>
+tree_node<T>* binary_tree<T>::insert_root(const T &data) {
+	return root = new tree_node<T>(data, nullptr, nullptr);
```
ssancho2: What if there was a root already?

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+template<class T>
+tree_node<T>* binary_tree<T>::insert_left(tree_node<T>* current,
```
ssancho2: It's a bit strange taking a node* as parameter, for example, it could be a node from another tree.
You could verify here that the tree doesn't already have a node with the same data.

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+			current->left = new tree_node<T>(data, nullptr, nullptr);
+		} else {
+			current->left->data = data;
+		}
+		return current->left;
+	}
+	return nullptr;
+}
+template<class T>
+tree_node<T>* binary_tree<T>::insert_right(tree_node<T>* current,
```
ssancho2: Maybe better to collapse this function and insert_left?
You will have a extra parameter to indicate left or right, could be a bool, or enum (preferred).

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+template<class T>
+bool binary_tree<T>::empty() {
+	return root == nullptr;
+}
+template<class T>
+void binary_tree<T>::clear() {
+	if (empty()) {
+		return;
+	}
+	stack<tree_node<T>*> st;
```
ssancho2: You could call recursively and get a stack for free :)

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+template<class T>
+bool binary_tree<T>::print_ancestors_helper(tree_node<T>* current,
+		const T &target) {
+	if (!current) {
+		return false;
+	}
+
+	if (print_ancestors_helper(current->left, target)) {
+		cout << current->data << ' ';
```
ssancho2: Please, make the function return the list of ancestors (in a `vector<T>` for example) and print it on main.

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+template<class T>
+tree_node<T>* binary_tree<T>::get_right(tree_node<T>* current) {
+	if (current) {
+		return current->right;
+	}
+	return nullptr;
+}
+template<class T>
+void binary_tree<T>::remove_leaf(tree_node<T>* current) {
+	if (current && !current->left && !current->right) {
```
ssancho2: What happens with the parent (left or right) pointing at this leaf?

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+	res_left.second = res_left.second + res_right.second;
+	if (res_left.second == 2) {
+		res_left.first = current;
+		return res_left;
+	}
+
+	return res_left;
+}
+template<class T>
+pair<tree_node<T>*, T> binary_tree<T>::common_ancestor(tree_node<T> *first,
```
ssancho2: Returning a pair here is not needed, right?
You could just return either a `tree_node<T>*` or a `T *`

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+class binary_tree;
+
+template<class T>
+class tree_node {
+	T data;
+	tree_node *left, *right;
+public:
+	tree_node(const T &data, tree_node* left, tree_node* right) :
+			data(data), left(left), right(right) {
+	}
```
ssancho2: You could have here public functions to get the value, left and right.

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+template<class T>
+void binary_tree<T>::print_ancestors(const T &target) {
+	if (!print_ancestors_helper(root, target)) {
+		cout << "ERROR: key not found in tree";
+	}
+	cout << endl;
+}
+template<class T>
+pair<tree_node<T>*, int> binary_tree<T>::common_ancestor_helper(
```
ssancho2: Writing in a comment of what the return is will be nice :)

MayssaraOmar/assignment2/Assignment 2.cpp
```diff
@@ -0,0 +1,275 @@
...
+public:
+	tree_node(const T &data, tree_node* left, tree_node* right) :
+			data(data), left(left), right(right) {
+	}
+	friend class binary_tree<T> ;
+// the binary_tree class is declared as tree_node's friend which means that binary_tree can access the private members of tree_node
+};
+
+template<class T>
+class binary_tree {
```
ssancho2: A note on design. It's not clear the separation of node and tree. Since several of the methods here take a node*, and those pointers could be in any tree, o no tree at all.
I think one some of the advantages you can have from a tree class (that you are not using) are:
* Finding a node by key.
* Ensuring all keys in a tree are different.

PR #15, opened 2018-06-05
Lacrit 2

Lacrit:


Lacrit/assignment2/binary_tree.hpp
```diff
@@ -0,0 +1,85 @@
...
+		T data;
+		Node *left; 
+		Node *right;
+		friend class Btree<T>;
+};
+
+template <class T>
+class Btree 
+{
+	// I need to have a reference to a pointer, otherwise it would overide the pointer? (Please, correct me if I`m wrong here)
```
ssancho2: That is an option, but it's a bit confusing. 
I think is better to have a pointer to a pointer.
In any case, it will be better if there was a way to construct any tree you can think of, this can only create unbalanced trees, right?

Lacrit/assignment2/binary_tree.hpp
```diff
@@ -0,0 +1,85 @@
...
+	Node<T>* get_root() 
+	{
+		return root;
+	};
+	bool print_ancestors(std::vector<T>& res, Node<T> *node, T key)
+	{
+		if ( node == nullptr ) return false;
+		if ( node->data == key ) return true; 
+
+		int a = 0;
```
ssancho2: Casting bools to ints an adding is a bit confusing :)
Why not something `bool res = print_ancestors(res, node->left, key) || print_ancestors(res, node->right, key);`?

Lacrit/assignment2/binary_tree.hpp
```diff
@@ -0,0 +1,85 @@
...
+		insert(key, root);
+	};
+	Node<T>* get_root() 
+	{
+		return root;
+	};
+	bool print_ancestors(std::vector<T>& res, Node<T> *node, T key)
+	{
+		if ( node == nullptr ) return false;
+		if ( node->data == key ) return true;
```
ssancho2: NIT: When you find the key, it's a good moment to clear the input vector, no guarantees that it was empty to start with.

Lacrit/assignment2/binary_tree.hpp
```diff
@@ -0,0 +1,85 @@
...
+	};
+	void insert(T key)
+	{
+		insert(key, root);
+	};
+	Node<T>* get_root() 
+	{
+		return root;
+	};
+	bool print_ancestors(std::vector<T>& res, Node<T> *node, T key)
```
ssancho2: It'll be good to split this function in 2: A public one, without the node parameter, and this one made private.

Lacrit/assignment2/q1.cpp
```diff
@@ -0,0 +1,52 @@
...
+	// ==> ancestors of 6 are 7->4->5->1 in any order (I`m doing in order of insertion)
+	// question 1
+	std::vector<int> res; // a vector to store the ancestors (not to print in the function) 
+	btree->print_ancestors(res, btree->get_root(), 6);
+	for ( size_t i = 0; i < res.size(); i++ )
+		if ( i == res.size()-1 ) std::cout << res.at(i) << std::endl; 
+		else std::cout << res.at(i) << "->";
+
+	// question 2 
+	Node<int> *lca = btree->least_common_ancestor(btree->get_root(), 2, 6);
```
ssancho2: Where is the implementation of this?

Lacrit:
Oh, I didn`t update the actual class with the method. My bad, sorry.
1)Why should I prefer a pointer to a pointer?
2)You can still construct it by calling insert on a parent and then on
children
3)Alright, I`ll change it
4)=//=
5)=//=
6)Uploading right now

2018-06-06 12:59 GMT+02:00 Sergio Sancho <notifications@github.com>:

Lacrit/assignment2/binary_tree.hpp
```diff
@@ -78,8 +81,22 @@ class Btree
 			return true;
 		}
 		return false;
-	}
+	};
+ 	Node<T> * least_common_ancestor(Node<T> *root, T val1, T val2)
+    {
+		// TODO: check if values are in the tree
```
ssancho2: Please keep identation consistent.

Lacrit/assignment2/binary_tree.hpp
```diff
@@ -78,8 +81,22 @@ class Btree
...
+	    if ( root == nullptr ) return nullptr;
+        if ( root->data == val1 || root->data == val2 ) return root;
+
+        Node<T> *left = least_common_ancestor(root->left, val1, val2);
+        Node<T> *right = least_common_ancestor(root->right, val1, val2);
+
+        if( left == nullptr && right == nullptr) return nullptr;
```
ssancho2: It's more clear this way, but this line is actually not needed :)

Lacrit:
I thought it actually should allow because that doesn`t contradict definition of a binary tree

ssancho2:
Sorry, what I mean is that the current code in the function `insert` won't allow that, but as you say that is a perfectly valid binary tree.

Lacrit:
should I open a new pull request for the changes?

ssancho2:
This pull request is fine, just update the files. The only issue remaining is the ability to create arbitrary trees, right?

PR #17, opened 2018-06-16
Create Solution.cpp

MennaMohie:
I'm not happy about the testing part, I'd really appreciate it if you could give me some advice.

### DIFF
--- a/MennaMohie/assignment3/Solution.cpp
+++ b/MennaMohie/assignment3/Solution.cpp
@@ -1,0 +1,172 @@
+#include<iostream>
+#include<string>
+#include<vector>
+#include<map>
+#include<unordered_map>
+#include<set>
+#include"Testing.h" //test macros + overloaded insertion operator for sets
+using namespace std;
+
+enum checkType { wordCheck, prefixCheck };
+
+const int gridCols = 3;
+
+class Trie;
+
+class TrieNode
+{
+	char character;
+	bool terminates = false;
+	unordered_map<char, TrieNode*> children;
+
+	void addWord(string word)
+	{
+		int wordSize = word.length();
+
+		if (wordSize == 0)
+			return;
+
+		char firstChar = word[0];
+		TrieNode *child = children[firstChar];
+		if (child == nullptr)
+		{
+			child = new TrieNode(firstChar);
+			children[firstChar] = child;
+		}
+		if(wordSize>1)
+			child->addWord(word.substr(1, wordSize - 1));
+		else
+			child->terminates = true;
+	}
+
+	bool hasChild(char c)
+	{
+		return children.find(c) != children.end();
+	}
+
+public:
+	TrieNode() {}
+	TrieNode(char c) :character(c){}
+	
+	friend class Trie;	
+};
+
+class Trie
+{
+	TrieNode *root;
+
+public:
+
+	Trie(vector<string> words)
+	{
+		root = new TrieNode();
+		for (string word : words)
+		{
+			root->addWord(word);
+		}
+	}
+
+	bool isWord(string word, checkType type = wordCheck)
+	{
+		if (word.length() == 0)
+			return false;
+
+		TrieNode *node = root;
+
+		for (char c : word)
+		{
+			if (!node->hasChild(c))
+				return false;
+			node = node->children[c];
+
+			//if we're doing a wordCheck and we've reached the last letter in the word we need to check its terminates boolean
+			if (type == wordCheck && c == word[word.length() - 1] && node->terminates == false)
+				return false;
+		}
+
+		return true;
+	}
+
+	bool isPrefix(string prefix)
+	{
+		return isWord(prefix, prefixCheck);
+	}
+};
+
+bool formWords(string word, Trie dictionary, char grid[][gridCols], int rows, int currentRow, int currentCol, vector<string> &words)
+{
+	if (!dictionary.isPrefix(word))
+		return false;
+
+	string updatedWord;
+	int updatedRow, updatedCol;
+	for (int x = -1; x <= 1; x++)
+	{
+		for (int y = -1; y <= 1; y++)
+		{
+			updatedRow = currentRow + x;
+			updatedCol = currentCol + y;
+			if ((y != 0 || x != 0) && updatedRow>=0 && updatedRow<rows && updatedCol>=0 && updatedCol<gridCols)
+			{
+				updatedWord = word + grid[currentRow + x][currentCol + y];
+				if (dictionary.isWord(updatedWord))
+				{
+					words.push_back(updatedWord);
+				}
+				formWords(updatedWord, dictionary, grid, rows, updatedRow, updatedCol, words);
+			}
+		}
+	}
+	return words.size() != 0;
+}
+
+set<string> formedWords(Trie dictionary, char grid[][3], int rows)
+{
+	set<string> allFormedWords;
+	vector<string> words;
+	string word;
+
+	for (int r = 0; r < rows; r++)
+	{
+		for (int c = 0; c < gridCols; c++)
+		{
+			words.clear();
+			word = grid[r][c];
+			if (formWords(word, dictionary, grid, rows, r, c, words))
+			{
+				for (string word : words)
+				{
+					allFormedWords.emplace(word);
+				}
+			}
+			
+		}
+	}
+	return allFormedWords;
+}
+
+int main()
+{
+	vector<string> v = { "CAR", "CARD", "CART", "CAT" };
+	Trie dict(v);
+
+	EXPECT_TRUE(dict.isPrefix("C"));
+	EXPECT_TRUE(dict.isPrefix("CAT"));
+	EXPECT_FALSE(dict.isPrefix(""));
+
+	EXPECT_TRUE(dict.isWord("CART"));
+	EXPECT_FALSE(dict.isWord("CA"));
+	EXPECT_FALSE(dict.isWord(""));
+
+	char arr[2][gridCols] = { 'A','A','R','T','C','D' };
+
+	set<string> actual = formedWords(dict, arr, 2);
+	set<string> expected = { "CAT", "CARD", "CAR" };	
+	EXPECT_EQ(expected, actual);
+	expected.clear();
+	expected = { "CAT", "CARD" };
+	EXPECT_EQ(expected, actual);
+	
+
+	return 0;
+}
--- a/MennaMohie/assignment3/Solution.cpp
+++ b/MennaMohie/assignment3/Solution.cpp
@@ -9,32 +9,30 @@ using namespace std;
 
 enum checkType { wordCheck, prefixCheck };
 
-const int gridCols = 3;
-
 class Trie;
 
 class TrieNode
 {
-	char character;
 	bool terminates = false;
 	unordered_map<char, TrieNode*> children;
 
-	void addWord(string word)
+	//passing the original word and the current index instead of making a different string for each call
+	void addWord(const string &word, int currentIndex=0)
 	{
-		int wordSize = word.length();
+		int wordSize = word.length()-currentIndex;
 
 		if (wordSize == 0)
 			return;
 
-		char firstChar = word[0];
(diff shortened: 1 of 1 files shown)


MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+enum checkType { wordCheck, prefixCheck };
+
+const int gridCols = 3;
+
+class Trie;
+
+class TrieNode
+{
+	char character;
```
ssancho2: Is this needed?

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+		char firstChar = word[0];
+		TrieNode *child = children[firstChar];
+		if (child == nullptr)
+		{
+			child = new TrieNode(firstChar);
+			children[firstChar] = child;
+		}
+		if(wordSize>1)
+			child->addWord(word.substr(1, wordSize - 1));
```
ssancho2: I think is fine this way, but be aware that this makes a lot of new strings as you go through a word, the other option is to pass around the original word and a index (is more confusing but it's faster).

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+		TrieNode *node = root;
+
+		for (char c : word)
+		{
+			if (!node->hasChild(c))
+				return false;
+			node = node->children[c];
+
+			//if we're doing a wordCheck and we've reached the last letter in the word we need to check its terminates boolean
+			if (type == wordCheck && c == word[word.length() - 1] && node->terminates == false)
```
ssancho2: `c == word[word.length() - 1]` doesn't work, you may have the last character repeated, for example checking `isWord('banana')` will fail in the first 'a', right?
It's easier to do the check after the loop, since the node variable points to the last node and you can check if that one is terminal or not.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+	Trie(vector<string> words)
+	{
+		root = new TrieNode();
+		for (string word : words)
+		{
+			root->addWord(word);
+		}
+	}
+
+	bool isWord(string word, checkType type = wordCheck)
```
ssancho2: It's a bit better to have 3 functions, 2 public: isWord & isPrefix, and the private implementation with the 'type' parameter.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+	{
+		if (word.length() == 0)
```
ssancho2: string has an empty() method.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include<vector>
+#include<map>
+#include<unordered_map>
+#include<set>
+#include"Testing.h" //test macros + overloaded insertion operator for sets
+using namespace std;
+
+enum checkType { wordCheck, prefixCheck };
+
+const int gridCols = 3;
```
ssancho2: This looks like a hack :)

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+		return true;
+	}
+
+	bool isPrefix(string prefix)
+	{
+		return isWord(prefix, prefixCheck);
+	}
+};
+
+bool formWords(string word, Trie dictionary, char grid[][gridCols], int rows, int currentRow, int currentCol, vector<string> &words)
```
ssancho2: On each call, you may be making a copy of all parameters, which can have performance implications.
Also, is a bit restrictive to allow boards of only 3 columns. May be better to have a `vector<vector<char>>` or `vector<string>`

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+		{
+			updatedRow = currentRow + x;
+			updatedCol = currentCol + y;
+			if ((y != 0 || x != 0) && updatedRow>=0 && updatedRow<rows && updatedCol>=0 && updatedCol<gridCols)
+			{
+				updatedWord = word + grid[currentRow + x][currentCol + y];
+				if (dictionary.isWord(updatedWord))
+				{
+					words.push_back(updatedWord);
+				}
```
ssancho2: You are not using the isPrefix function... What can it be useful for?

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+int main()
+{
+	vector<string> v = { "CAR", "CARD", "CART", "CAT" };
+	Trie dict(v);
+
+	EXPECT_TRUE(dict.isPrefix("C"));
+	EXPECT_TRUE(dict.isPrefix("CAT"));
+	EXPECT_FALSE(dict.isPrefix(""));
```
ssancho2: Well, this one is arguable... All words kind of start with "" :)

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include<unordered_map>
+#include<set>
+#include"Testing.h" //test macros + overloaded insertion operator for sets
+using namespace std;
+
+enum checkType { wordCheck, prefixCheck };
+
+const int gridCols = 3;
+
+class Trie;
```
ssancho2: You made a Trie! Nice!!!

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+	bool hasChild(char c)
+	{
+		return children.find(c) != children.end();
+	}
+
+public:
+	TrieNode() {}
+	TrieNode(char c) :character(c){}
+
```
ssancho2: Not a big deal, but you may need a destructor to free all nodes created.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+	char arr[2][gridCols] = { 'A','A','R','T','C','D' };
+
+	set<string> actual = formedWords(dict, arr, 2);
+	set<string> expected = { "CAT", "CARD", "CAR" };	
+	EXPECT_EQ(expected, actual);
+	expected.clear();
+	expected = { "CAT", "CARD" };
+	EXPECT_EQ(expected, actual);
+
```
ssancho2: Regarding test, not much more to add. 
A few more tests for dictionary, some prefixes that are not there. Some lower case letters.
Maybe test with a empty dictionary.
For the actual method maybe add another board/dictionary.

MennaMohie/assignment3/Solution.cpp
MennaMohie: After the if-condition, the formWords function is called with the new word, and first thing it does is to check isPrefix, so I am using it...
Or do you mean something else?

MennaMohie/assignment3/Solution.cpp
```diff
@@ -66,107 +95,171 @@ class Trie
 		}
 	}
 
-	bool isWord(string word, checkType type = wordCheck)
+	//I need help doing the copy constructor :/
```
ssancho2: Well, I actually don't think you actually need a copy constructor. I mean, it may be useful to be able to do a (deep) copy of Trie, but it's well beyond the current problem. The Finder class can keep a pointer to a Trie, instead of a copy of it (not the best design, it'll be a bit better to have a unique_ptr, but don't get into that if you don't want :) ) 
To be sure the compiler is not trying to copy nodes, you can add:
Trie(const Trie&) = delete;
Trie& operator =(const Trie&) = delete;
If you do want to do a copy, you need a deep one, i.e, you need to recursively make a copy of every node in the Trie, keeping the structure, of course.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -45,16 +43,47 @@ class TrieNode
...
+		{
+			unordered_map<char, TrieNode*>::iterator it;
+			for (it = node->children.begin(); it != node->children.end(); it++)
+				remove(it->second);
+		}
+	}
+
 public:
 
 	Trie(vector<string> words)
```
ssancho2: Better to make the parameter `const vector<string>&` to avoid a copy

MennaMohie/assignment3/Solution.cpp
ssancho2: My bad, you are right, didn't see it :)

MennaMohie/assignment3/Solution.cpp
```diff
@@ -66,107 +95,171 @@ class Trie
...
 
-set<string> formedWords(Trie dictionary, char grid[][3], int rows)
-{
-	set<string> allFormedWords;
-	vector<string> words;
-	string word;
+public:
 
-	for (int r = 0; r < rows; r++)
+	Finder(const Trie &d, const vector<vector<char>> &g, const int r, const int c) :dictionary(d), grid(g), rows(r), cols(c) {}
```
ssancho2: r & c parameters are not needed, since you can check the size of the grid directly from g.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -66,107 +95,171 @@ class Trie
...
+	grid2.push_back({ 'l','g','o','n' });
+	grid2.push_back({ 'e','n','a','d' });
+	grid2.push_back({ 'm','a','c','b' });
+
+	Finder finder2(dict2, grid2, 5, 4);
+
+	set<string> actual2 = finder2.formedWords();
+	set<string> expected2 = { "banana", "apple", "mango" };
+	EXPECT_EQ(expected2, actual2);
+	expected2.clear();
```
ssancho2: Not needed.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -66,107 +95,171 @@ class Trie
...
+	expected2 = { "Banana", "apple", "mango" };
+	EXPECT_EQ(expected2, actual2);
```
ssancho2: Does this work?

MennaMohie/assignment3/Solution.cpp
```diff
@@ -66,107 +95,171 @@ class Trie
...
-			if (formWords(word, dictionary, grid, rows, r, c, words))
+			for (int c = 0; c < cols; c++)
 			{
-				for (string word : words)
+				visited.clear();
+				visited.resize(rows,vector<bool>(cols,false));
+				
+				words.clear();
+				word = grid[r][c];
+				if (formWords(word, r, c, words))
```
ssancho2: Better to pass allFormedWords to formWords and let it keep adding to it, right?
You avoid having to copy them.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -66,107 +95,171 @@ class Trie
...
-			updatedRow = currentRow + x;
-			updatedCol = currentCol + y;
-			if ((y != 0 || x != 0) && updatedRow>=0 && updatedRow<rows && updatedCol>=0 && updatedCol<gridCols)
+			for (int y = -1; y <= 1; y++)
 			{
-				updatedWord = word + grid[currentRow + x][currentCol + y];
-				if (dictionary.isWord(updatedWord))
+				updatedRow = currentRow + x;
+				updatedCol = currentCol + y;
+				if ((y != 0 || x != 0) && updatedRow >= 0 && updatedRow < rows && updatedCol >= 0 && updatedCol < cols)
```
ssancho2: Minor comment, `(y != 0 || x != 0)` is not needed, since the current cell is marked visited already.

MennaMohie/assignment3/Solution.cpp
```diff
@@ -66,107 +95,171 @@ class Trie
...
 				{
-					words.push_back(updatedWord);
+					if (!visited[updatedRow][updatedCol])
```
ssancho2: This if can be collapsed with the previous one.

PR #18, opened 2018-06-18
JoodHajeer-3

JoodHajeer:


JoodHajeer/assignment3/JoodHajeer-3.cpp
```diff
@@ -0,0 +1,234 @@
...
+	findWords(dictionary_, charGrid, words);
+	EXPECT_EQ(0, words.size());
+}
+void testFindWords_gridOneCell(Dictionary & dictionary_, vector<vector<char> > & charGrid, set<string> & words) {
+	charGrid.clear();
+	testFindWords_fillCharGrid(1, 1, "C", charGrid);
+	findWords(dictionary_, charGrid, words);
+	EXPECT_EQ(0, words.size());
+}
+int main() {
```
ssancho2: Some comments about testing:
It's better if the functions to do each test are encapsulated, and there is no state flowing in/out of them.
On thing you can take advantage of is c++ list-initialization. For example, if you have a function:
 `void doOneTest(const vector<string> &grid, const set<string> &dictionary, ...) `
You can call it like:
`doOneTest({"AAR", "TCD"}, {"CAT", "CAR", "CART", "CARD" }, ...`

ssancho2: Sure, no problem.

PR #19, opened 2018-06-19
MayssaraOmar-3 (add assignment 3 solution)

MayssaraOmar:


MayssaraOmar/assignment3/Word search/.cproject
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
flerdacodeu: It seems you accidentally added a number of files that are configuration files for your IDE.
I would suggest you delete those.

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -0,0 +1,48 @@
+#include <map>
+#include "word search.h"
+using namespace std;
+
+void get_words_from_grid(vector<vector<char> > &grid, int row, int col,
+		string word, map<pair<int, int>, bool> &vis,
```
flerdacodeu: Optional: I would avoid shortening variable names like "visited" to "vis", because it makes the code harder to read.

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -0,0 +1,48 @@
...
+		map<pair<pair<int, int>, string>, bool> &memo, dictionary &dic,
```
flerdacodeu: "memo" is not very clear in this context: what is this supposed to represent?
You can add this as a comment to the function.

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -0,0 +1,48 @@
...
+		unordered_set<string> &words) {
```
flerdacodeu: Style: A common convention is to make output parameters (parameters that are outputs or modified by a function) pointers rather than references.

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -0,0 +1,48 @@
...
+	if (row < 0 || row >= grid.size() || col < 0 || col >= grid[row].size()) {
+		return;
+	}
+	if (!dic.isPrefix(word + grid[row][col])
+			|| memo[make_pair(make_pair(row, col), word)]) {
+		return;
+	}
+
+	if (dic.isWord(word + grid[row][col])) {
```
flerdacodeu: This depends on exactly how `isPrefix()` works, but it is possible it would not return true for a completed word that is a prefix of another word. In that case, I would suggest checked `isWord()` first.

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -0,0 +1,48 @@
...
+						word + grid[row][col], vis, memo, dic, words);
+				// (row-1 col-1 left up) (row-1 col up) (row-1 col+1 up right)
+				// (row col-1 left) (row col current) (row col+1 right)
+				// (row+1 col-1 left down) (row+1 col down) (row+1 col+1 down right)
+				vis[make_pair(row + i, col + j)] = false;
+			}
+		}
+	}
+	memo[make_pair(make_pair(row, col), word)] = true;
+
```
flerdacodeu: Style: remove this blank line, since it does not particularly seems to clarify the code or help understand its structure. In fact, I would probably put a blank line between the functions instead, since there it is more meaningful (help visualise that one function is ended and the next one is starting).

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+void get_words_from_grid(vector<vector<char> > &grid, dictionary &dic,
```
flerdacodeu: I would actually put a comment on the previous function as being internal, while this is what you would expect people to call. This is often done by making the above function `static`, which means it can only be accessed by this file. This function is already declared in a `.h` file, making it clear it is meant to be used from elsewhere.

MayssaraOmar/assignment3/Word search/src/Word search.h
```diff
@@ -0,0 +1,11 @@
+
```
flerdacodeu: Remove these blank lines at the top of the file.

MayssaraOmar/assignment3/Word search/src/Word search.h
```diff
@@ -0,0 +1,11 @@
...
+
+#ifndef WORD_SEARCH_H_
+#define WORD_SEARCH_H_
+
+#include <unordered_set>
+#include "dictionary.h"
+
+void get_words_from_grid(vector<vector<char> > &grid, dictionary &dic,unordered_set<string> &ans);
```
flerdacodeu: A comment here exampling what the function does and what it expects as arguments would be useful for someone wanting to use this code.

flerdacodeu: Style: make `grid` and `dic` const references, because you should not be modifying the grid or the dictionary while searching for words.

MayssaraOmar/assignment3/Word search/src/dictionary.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+	node* ptr = root.get();
+	for (char c : word) {
+		if (!ptr->children[c]) {
+			ptr->children[c] = unique_ptr<node>(new node);
+		}
+		ptr = ptr->children[c].get();
+	}
+	ptr->isEndWord = true;
+	return;
```
flerdacodeu: Optional. It is okay to have a return at the end of the function, but it is actually not required, so I would suggest leaving it out.

MayssaraOmar/assignment3/Word search/src/dictionary.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+void dictionary::insert(const vector<string> &words) {
+
+	for (string word : words) {
+		insert(word);
+	}
+	return;
+}
+
+void dictionary::remove(const string &word) {
```
flerdacodeu: You do not actually need `remove`, so you might as well leave it out.

MayssaraOmar/assignment3/Word search/src/dictionary.cpp
```diff
@@ -0,0 +1,84 @@
...
+bool dictionary::isPrefix(const string &word) {
+
+	node* ptr = root.get();
+	for (char c : word) {
+		if (!ptr->children[c]) {
+			return false;
+		}
+		ptr = ptr->children[c].get();
+	}
+	return true;
```
flerdacodeu: If you wanted to implement a strict prefix check, you could check if the final node has any children.
It would be nice to document the behavior either way in your `.h` file.

MayssaraOmar/assignment3/Word search/src/dictionary.h
```diff
@@ -0,0 +1,40 @@
...
+		bool isEndWord;
+		node() {
+			isEndWord = false;
+		}
+	};
+	unique_ptr<node> root;
+
+public:
+	dictionary();
+	//~dictionary(); no need for a destructor as we use smart pointers right?
```
flerdacodeu: That is correct!
The other reason you might need a destruction is if this class has any virtual function (which in this case it does not).

MayssaraOmar/assignment3/Word search/src/dictionary.h
```diff
@@ -0,0 +1,40 @@
...
+#include <vector>
+#include <string>
+#include <cstdlib>
+using namespace std;
+
+
+#ifndef DICTIONARY_H_
+#define DICTIONARY_H_
+
+class dictionary {
```
flerdacodeu: It would be nice to add a few comments to this class declaration.

MayssaraOmar/assignment3/Word search/src/main.cpp
```diff
@@ -0,0 +1,14 @@
+//============================================================================
+//ASSUMPTIONS
+// (dictionary) class and (get_words_from_grid) method are case sensitive
+//============================================================================
+
+#include <iostream>
+#include "test_dictionary.h"
+#include "test_wordSearch.h"
```
flerdacodeu: Optional: it is often the case that tests are put in the same file as the main function that invoked them.

MayssaraOmar/assignment3/Word search/src/test_dictionary.cpp
```diff
@@ -0,0 +1,93 @@
...
+	remove_vector();
+	isWord_true();
+	isWord_false();
+	isPrefix_true();
+	isPrefix_false();
+	clear_empty();
+	empty_dictionary();
+}
+void test_dictionary::insert_oneword() {
+	dic.insert("hello");
```
flerdacodeu: It is often the case that in your test you want to check that things behaved correctly.
For example, here you could say:

```c++
   dic.insert("hello");
   EXPECT_TRUE(dic.isWord("hello"));
   EXPECT_TRUE(dic.isPrefix("h"));
   EXPECT_TRUE(dic.isPrefix("he"));
   EXPECT_TRUE(dic.isPrefix("hel"));
   EXPECT_TRUE(dic.isPrefix("hell"));
   EXPECT_TRUE(dic.isPrefix("hello"));
```

Otherwise your test runs the code but it does not check that things worked.
This shows how it is not always possible to test functions completely in isolation.

MayssaraOmar/assignment3/Word search/src/test_dictionary.cpp
```diff
@@ -0,0 +1,93 @@
+#include "test_dictionary.h"
+#include "test.h"
+#include "exception"
+
+dictionary test_dictionary::dic;
+
+void test_dictionary::run_all_tests() {
```
flerdacodeu: One problem with the current test setup is that all tests share the same dictionary, which means the order you run the tests has an effect on the result.
An alternative would be to make the test class have non-static methods and for each test you create an instance.

It would then look something like:

```c++
void test_dictionary::run_all_tests() {
  new test_directionary().insert_oneword();
  new test_directionary().insert_vector();
  ...
}
```

MayssaraOmar/assignment3/Word search/src/test_dictionary.h
```diff
@@ -0,0 +1,22 @@
+#ifndef TEST_DICTIONARY_H_
+#define TEST_DICTIONARY_H_
+
+#include "dictionary.h"
+
+class test_dictionary {
```
flerdacodeu: Style: Generally classes are named in CamelCase, so I would suggest `DictionaryTests` for this class.

MayssaraOmar/assignment3/Word search/src/test_wordSearch.cpp
```diff
@@ -0,0 +1,136 @@
...
+dictionary test_wordSearch::dic;
+
+void test_wordSearch::setup() {
+	dic.insert("car");
+	dic.insert("card");
+	dic.insert("cart");
+	dic.insert("cat");
+	dic.insert("m");
+}
+void test_wordSearch::run_all_tests() {
```
flerdacodeu: Similarly to above, it would make sense to run all the tests on a different instance to avoid interactions between them.

Since you have a setup method you could:
* call setup() in the constructor
* call setup() explicitly for each test

For the latter on, you could use:

```c++
test_wordSearch& test_WordSearch::setup() {
  ...
  return *this;
}

...


void test_wordSearch::run_all_tests() {
  new test_wordSearch().setup().empty_grid();
  new test_wordSearch().setup().one_cell_grid_with_match_in_dictionary();
  ...
}
```

Does this make sense?

flerdacodeu:
Really sorry for the delay on this! I had messed up with my comments.

In terms of the use of smart points to manage your data structure, I think you did a great job.

The main advantage of smart pointers (in my opinion) is that you avoid accidentally forgetting to delete things.

In this specific case, you could simply ignore the destructor (which could have ended up as a complex piece of code traversing the data structure) and your `remove` methods are a bit simpler to.

The advantages are usually more evident where you have large data structures that you might need to clean up in non-trivial ways, if there are various ways to modify them, or in case of exceptional behavior. The last one, exceptional behavior, is something you did not have here, so let me give you more details.

Let's assume that you need to allocate an object for your function: you might end up storing the object by the end of the function, but you might also return early or throw an exception; if you do have this behavior (something that happens only on special cases), then you need to free allocated object in these special code paths, but it is really easy to forget to do so.

```c++
bool findWords(const Grid& grid, vector<string>* words) {
  Dictionary* dic = new Dictionary();
  if (!loadDictionary(dic)) {
    return false;
  }
  return findWordsInDirectionary(grid, *dic, words);
}
```

Now, if you cannot load the dictionary object, you will *leak* that object, as you forgot to call delete. The same code with a smart pointer will work correctly by deleting the object. This is not a perfect example (you could have had the dictionary instead of a pointer to id), but in some cases it is not always possible to do so.

Essentially smart pointers are a way to simplify your code, as the smart pointer takes care of many of the things that are needed for you.

MayssaraOmar/assignment3/Word search/src/test_wordSearch.cpp
MayssaraOmar: This is what I understood:

1- If I call setup() in the constructor, then all tests still share the same dictionary and there are interactions between tests(unless I'm making a new instance for each test). 

2- I don't need to make  a different instance for each test if each test is completely independent (has its own dictionary), for example calling  setup(dictionary*) explicitly for each test with its own dictionary.

I think method 2 is simpler as I don't have the responsibility of freeing the memory I allocated for each test.

MayssaraOmar/assignment3/Word search/src/Word search.cpp
MayssaraOmar: For example the following grid with the word "match" in dictionary:
a a r
t m d
c h x
We can obtain the word "match" by 2 ways:
1- by going from 'm' to 'a' upwards and then 't', 'c', 'h'
2- by going from 'm' to 'a' diagonally and then 't', 'c', 'h'
Here the ('t', 'c', 'h') is the repeated work as we call the function with the same input (row = 1, col = 0, word = "ca") twice.
We avoid repeating work by marking (1,0,"ca") true, in the second call, (1,0,"ca") will be true and the function will return.

MayssaraOmar/assignment3/Word search/src/WordSearchTest.cpp
```diff
@@ -23,114 +22,169 @@ void test_wordSearch::run_all_tests() {
...
 }
-void test_wordSearch::case_sensitive() {
-	vector<vector<char> > grid(3);
 
+void WordSearchTest::case_sensitive() {
+	// setup
+	dictionary dic;
+	setup(&dic);
+	// build grid
+	vector<vector<char> > grid(3);
```
ssancho2: This could be done like:
`vector<vector<char> > grid = { {'a', 'a', 'r'}, {'t', 'c', 'D'}} ;`

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -2,47 +2,54 @@
...
-void get_words_from_grid(vector<vector<char> > &grid, int row, int col,
-		string word, map<pair<int, int>, bool> &vis,
-		map<pair<pair<int, int>, string>, bool> &memo, dictionary &dic,
-		unordered_set<string> &words) {
+// internal function
+// cache maps a pair of a (cell position (row, col)) and a string (prefix) to a boolean to ensure that (get_words_from_grid) method doesn't run for the same inputs more than once using memoization technique
+static void get_words_from_grid(const vector<vector<char> > &grid, int row, int col,
+		string prefix, map<pair<int, int>, bool>* visited,
+		map<pair<pair<int, int>, string>, bool>* cache, const dictionary &dic,
+		unordered_set<string>* words) {
```
ssancho2: Some comments on the function / parameters:
I like the cache, avoiding checking twice the same thing, but why do you think is important? How likely it's that you will hit it? 
I think the visited parameter will be slightly better (more performant) as a vector<vector<bool>>

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -2,47 +2,54 @@
...
+	// prefix doesn't exist in dictionary
+	if (!dic.isPrefix(prefix + grid[row][col])) {
+		return;
 	}
 	// recurse on all non-visited 8 directions
 	for (int i = -1; i <= 1; i++) {
 		for (int j = -1; j <= 1; j++) {
-			if (!vis[make_pair(row + i, col + j)]) {
-				vis[make_pair(row + i, col + j)] = true;
+			if (!(*visited)[make_pair(row + i, col + j)]) {
```
ssancho2: Maybe you can keep `make_pair(row + i, col + j)` in a variable.

MayssaraOmar/assignment3/Word search/src/Word search.cpp
```diff
@@ -2,47 +2,54 @@
...
+	// result has been already computed for these inputs
+	if((*cache)[make_pair(make_pair(row, col), prefix)]){
 		return;
 	}
-
-	if (dic.isWord(word + grid[row][col])) {
-		words.insert(word + grid[row][col]);
+	// word exits in dictionary
+	if (dic.isWord(prefix + grid[row][col])) {
+			words->insert(prefix + grid[row][col]);
```
ssancho2: You are doing `prefix + grid[row][col]` in several places, may be easier to do `prefix += grid[row][col];` before this if

MayssaraOmar/assignment3/Word search/src/Word search.cpp
MayssaraOmar: I think it depends on the type of inputs, if there is a lot of duplicates nearby in the grid, then the cache might be useful to optimize time. However, in all other types of input, it reserves unnecessary memory.

MayssaraOmar/assignment3/Word search/src/WordSearchTest.cpp
MayssaraOmar: my code doesn't compile with this kind of vector initialization :/

MayssaraOmar: My code doesn't compile with this kind of vector initialization :/ I'll leave it as it is for now and I'll look into the problem

PR #20, opened 2018-06-19
Add files via upload

marahgh:


marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
+#include <iostream>
+#include <vector>
+#include <algorithm>
+using namespace std;
+
+/*A helper function:returns true if the given word
+was found in the given vector, and false otherwise*/
+bool isFoundInVector(vector<string> words, string word) {
+  if (std::find(words.begin(), words.end(), word) != words.end()) {
```
flerdacodeu: Whenever you see the pattern:

```c++
  if (...) {
   return true;
  } else {
   return false;
  }
```

You should be able to replace it with a single `return`-statement.

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
...
+/*A helper function:returns true if the given word
+was found in the given vector, and false otherwise*/
+bool isFoundInVector(vector<string> words, string word) {
+  if (std::find(words.begin(), words.end(), word) != words.end()) {
+    return true;
+  }
+  return false;
+}
+
+/*The grid class,
```
flerdacodeu: Optional: This comment seems to be split in the middle...

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
...
+  int rows;
+  int cols;
+  bool empty;
+
+ public:
+  /*creates an empty grid,
+  a grid with the given dimensions filled by ' ' charachters*/
+  Grid(int rows, int cols)
+      : matrix(nullptr), rows(rows), cols(cols), empty(true) {
+    matrix = new char*[rows];
```
flerdacodeu: There seems to be some duplication between the two constructors.
What you might be able to do is use a delegating constructor:

```c++
  Grid(int rows, int cols) : Grid(vector<char>(rows * cols, ' ')) {}
```

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
...
+  return false;
+}
+
+/*The grid class,
+contains lower-case letters from whichever alphabet the dictionary uses.*/
+class Grid {
+  char** matrix;
+  int rows;
+  int cols;
+  bool empty;
```
flerdacodeu: For things that are now immediately clear (like `empty`), you can add a comment describing what this member variable is for.

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
...
+      return false;
+    }
+    for (int i = 0; i < rows; i++) {
+      for (int j = 0; j < cols; j++) {
+        matrix[i][j] = chars[(i * cols) + j];
+      }
+    }
+    return true;
+  }
+  /*Grid's destructor*/
```
flerdacodeu: This comment does not add much useful information, so I would suggest to just leave it out.

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
...
+    }
+    return true;
+  }
+
+  /*Removes hyphens("-") and converts letters to lower case(returns a legal
+  dictionary word)
+  if the word contains alphabet letters and hyphens, other wise returns an empty
+  word */
+  string fixWord(string word) {
+    string fixed = "";
```
flerdacodeu: Constructing a new string is actually a bit inefficient. We could instead modify the one we were given (since it is a copy anyway).

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
...
+  /*creates an empty dictionary*/
+  Dictionary() { dictionary = vector<string>(); }
+
+  /*Returns whether the given string is a valid word*/
+  bool isWord(string word) {
+    return isFoundInVector(dictionary, fixWord(word));
+  }
+
+  /*Adds a new non- empty word to the dictionary*/
+  bool addWord(string word) {
```
flerdacodeu: Most of the time, you want to use `const string& word` when accepting a string parameter.
This makes it so that no copy is created when the string is passed in.

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
...
+/*Dictionary class.
+Dictionary words are lowercase alphabet words*/
+class Dictionary {
+  vector<string> dictionary;
+
+  /****************************************************************************/
+  /* a private auxiliary method of isPrefix,
+  checks whether w2 is a prefix of w1*/
+  bool isWordPrefix(string w1, string w2) {
+    if (w1.size() < w2.size()) {
```
flerdacodeu: You can use `string::compare(pos, len, str)` to compare against a prefix instead.
When there is a function in the standard library that can be used to do what you need, there are usually advantages to using it (like it can be implemented more efficiently than an explicit loop).

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
...
+    word.pop_back();
+    return;
+  }
+  if (dictionary.isWord(word)) {
+    if (!isFoundInVector(words, word)) {
+      words.push_back(word);
+    }
+  }
+  char c = grid[i][j];
+  grid[i][j] = '*';
```
flerdacodeu: It would be nicer to be able to do this without modifying `grid` along the way.
Generally modifying the data structure you are working on is not ideal: for example, if you wanted to search for words in different languages in parallel, this code would not work correctly as the two different runs would overwrite each other's grid.

marahgh/assignment3/Testing.cpp
```diff
@@ -0,0 +1,137 @@
...
+  for (string word : words) {
+    dictionary.addWord(word);
+  }
+  return dictionary;
+}
+
+/***************************Ivnalid Inputs*************************************/
+/*Empty Dictionary*/
+
+Grid testEmptyDictionaryCreateGrid() {
```
flerdacodeu: It actually would be clearer to just create this as part of your test itself.

marahgh/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,212 @@
+#include <iostream>
```
flerdacodeu: It would make code more re-usable, if we split it into a `WordSearch.h` and `WordSearch.cpp` pair of files.
In fact, you might consider making the grid and dictionary their own pair of files.
This means you can reuse these more easily.
Also, at the moment, most of your methods are defined inline within the class declaration: this makes it harder to understand the structure of the class.

For example:

`grid.h`
```c++
/*The grid class, contains lower-case letters from whichever alphabet the dictionary uses.*/
class Grid {
  char** matrix;
  int rows;
  int cols;
  bool empty;

 public:
  /*creates an empty grid,
  a grid with the given dimensions filled by ' ' charachters*/
  Grid(int rows, int cols);
  /*receives a vector of the charachters to to fill the grid with,
  grid's rows number and grid's columns number.
  creates the grid if the dimensions are OK , otherwise creats an empty grid*/
  Grid(vector<char> chars, int rows, int cols);
  /*Fills the grid with the given values*/
  bool fillGrid(vector<char> chars);
  /*Grid's destructor*/
  ~Grid();
  /*returns the grid's rows number*/
  int getRows() { return rows; }
  /*returns the grid's columns number*/
  int getCols() { return cols; }
  /*return true if the grid is empty, and false otherwise*/
  bool isEmpty() { return empty; }
  char* operator[](int i);
};
```

`grid.cpp`
```c++
#include "grid.h"

Grid::Grid(int rows, int cols)
    : matrix(nullptr), rows(rows), cols(cols), empty(true) {
  matrix = new char*[rows];
  for (int i = 0; i < rows; i++) {
    matrix[i] = new char[cols];
  }
  vector<char> chars = vector<char>((rows * cols), ' ');
  empty = true;
  fillGrid(chars);
}

...
```

Looking at this, it also seems to me that `fillChars()` should not be public (which was not obvious in the previous form).

marahgh/assignment3/wordSearch.cpp
```diff
@@ -1,174 +1,17 @@
...
-    return false;
-  }
-};
-
 /*An auxiliary function of find word, receives a grid,
 cell cordinates to start searchnig a word from, a dictionary.
 returns the legal words starts form cell i,j int the words vector*/
-void findWord(Grid& grid, int i, int j, Dictionary dictionary, string word,
-              vector<string>& words) {
+void findWord(Grid& grid, Grid& visited, int i, int j, Dictionary dictionary,
```
ssancho2: Reusing the Grid class for visited it's a bit of a stretch :)
Nevertheless, you could use 0 & 1 (instead of '0' & '1') to make the code a bit more readable, since you can do things like `if (visited[i][j]) { ... } ` (you could even use true and false, but that is also a stretch :) )

marahgh/assignment3/wordSearch.cpp
```diff
@@ -181,17 +24,17 @@ void findWord(Grid& grid, int i, int j, Dictionary dictionary, string word,
...
-  findWord(grid, i - 1, j - 1, dictionary, word, words);
-  findWord(grid, i - 1, j, dictionary, word, words);
-  findWord(grid, i - 1, j + 1, dictionary, word, words);
-  findWord(grid, i + 1, j - 1, dictionary, word, words);
-  findWord(grid, i + 1, j, dictionary, word, words);
-  findWord(grid, i + 1, j + 1, dictionary, word, words);
-  grid[i][j] = c;
+
+  visited[i][j] = '1';
+  findWord(grid, visited, i, j - 1, dictionary, word, words);
```
ssancho2: Two nested for loops will be more readable.

marahgh/assignment3/Grid.cpp
```diff
@@ -0,0 +1,71 @@
+#include "Grid.h"
```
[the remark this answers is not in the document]
marahgh: actually this is a problem that i have since in the university i learn at teach us to implement almost everything we use so it didn't come to my mind to use it, this is a problem i'm getting better in conquering it because of these assignments :P thank you! :D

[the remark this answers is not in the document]
ssancho2: Again, I'm not saying it's wrong, there are arguments both ways. Just want you to think what are the advantages are of one vs the other.
No problem, happy to help :)

PR #21, opened 2018-06-20
Laiaaa0 3

laiaaa0:
Sorry for the delay!

### DIFF
--- a/laiaaa0/assignment3/Dictionary.cpp
+++ b/laiaaa0/assignment3/Dictionary.cpp
@@ -11,22 +11,60 @@ bool Dictionary::IsPrefixOfWord(std::string prefix, std::string word){
 }
 
 
-Dictionary::Dictionary(std::vector<std::string> &word_list){
-  this->dictionary_words_ = word_list;
+Dictionary::Dictionary(const std::vector<std::string> &word_list){
+  for (int i = 0; i<word_list.size(); ++i){
+    std::string current_word = word_list[i];
+    InsertWord(current_word);
+  }
 
 }
-bool Dictionary::IsWord (std::string query_word){
-  for (int i = 0; i<this->dictionary_words_.size(); ++i){
-    if (dictionary_words_[i] ==query_word) return true;
+bool Dictionary::IsWord (const std::string & query_word){
+  if (this->dictionary_words_.find(query_word) == this->dictionary_words_.end()) return false;
+  else return (this->dictionary_words_[query_word]);
+}
+
+
+bool Dictionary::IsPrefix (const std::string & query_prefix){
+  if (this->dictionary_prefixes_.find(query_prefix) == this->dictionary_prefixes_.end()) return false;
+  else {
+    if (this->dictionary_prefixes_[query_prefix]>0) return true;
+    else return false;
   }
-  return false;
 
 }
-bool Dictionary::IsPrefix (std::string query_prefix){
-  for (int i = 0; i<this->dictionary_words_.size(); ++i){
-    if (IsPrefixOfWord(query_prefix,dictionary_words_[i])) return true;
-  }
-  return false;
+bool Dictionary::InsertWord(const std::string & word){
+  //word already exists
+  if (this->dictionary_words_.find(word) != this->dictionary_words_.end()) return false;
+  else {//word doesnt exist
 
+    this->dictionary_words_[word] =true;
+    std::string prefix = "";
+    for (int j = 0; j<word.size(); ++j){
+        prefix = prefix+word[j];
+        if ( this->dictionary_prefixes_.find(prefix) == this->dictionary_prefixes_.end() ) {
+            this->dictionary_prefixes_[prefix] = 1;
+        }
+        else {
+          this->dictionary_prefixes_[prefix]++;
+        }
+
+    }
+    return true;
+  }
+}
+bool Dictionary::RemoveWord(const std::string & word){
+  //word doesnt exist
+  if (this->dictionary_words_.find(word) == this->dictionary_words_.end()) return false;
+  else { // word exists
+    this->dictionary_words_[word] = false;
+    std::string prefix = "";
+    for (int j = 0; j<word.size(); ++j){
+        prefix = prefix+word[j];
+        this->dictionary_prefixes_[prefix]--;
+
+    }
+
+    return true;
+  }
 
 }
--- a/laiaaa0/assignment3/Dictionary.cpp
+++ b/laiaaa0/assignment3/Dictionary.cpp
@@ -1,6 +1,6 @@
 #include "Dictionary.h"
 
-bool Dictionary::IsPrefixOfWord(std::string prefix, std::string word) {
+bool Dictionary::IsPrefixOfWord(std::string prefix, std::string word) const{
   if (prefix.size() > word.size()) return false;
   for (int i = 0; i < prefix.size(); ++i) {
     if (word[i] != prefix[i]) return false;
@@ -15,19 +15,20 @@ Dictionary::Dictionary(const std::vector<std::string> &word_list) {
   }
 }
 
-bool Dictionary::IsWord (const std::string & query_word) const {
-  if (this->dictionary_words_.find(query_word) == this->dictionary_words_.end()) return false;
-  return (this->dictionary_words_[query_word]);
+bool Dictionary::IsWord(const std::string & query_word) const {
+  auto query_iterator =this->dictionary_words_.find(query_word);
+  if (query_iterator == this->dictionary_words_.end()) return false;
+
+  return query_iterator->second;
 }
 
 
-bool Dictionary::IsPrefix (const std::string & query_prefix) const {
-  if (this->dictionary_prefixes_.find(query_prefix) == this->dictionary_prefixes_.end()) return false;
-  else {
-    if (this->dictionary_prefixes_[query_prefix]>0) return true;
-    else return false;
-  }
+bool Dictionary::IsPrefix(const std::string & query_prefix) const {
+  auto query_iterator = this->dictionary_prefixes_.find(query_prefix);
+  if (query_iterator == this->dictionary_prefixes_.end()) return false;
 
+  if (query_iterator->second > 0) return true;
+  return false;
 }
 bool Dictionary::InsertWord(const std::string & word) {
   //word already exists
@@ -58,10 +59,7 @@ bool Dictionary::RemoveWord(const std::string & word) {
     for (int j = 0; j<word.size(); ++j){
         prefix = prefix+word[j];
         this->dictionary_prefixes_[prefix]--;
-
     }
-
     return true;
   }
-
 }
--- a/laiaaa0/assignment3/Dictionary.cpp
+++ b/laiaaa0/assignment3/Dictionary.cpp
@@ -1,0 +1,32 @@
+#include "Dictionary.h"
+
+bool Dictionary::IsPrefixOfWord(std::string prefix, std::string word){
+  if (prefix.size()>word.size()) return false;
+  else {
+    for (int i=0; i<prefix.size(); ++i){
+      if (word[i]!=prefix[i]) return false;
+    }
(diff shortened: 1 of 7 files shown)


laiaaa0/assignment3/q1.cpp
```diff
@@ -0,0 +1,51 @@
+
+#include "TestingFunctions.cpp"
+
+int directions_x []= {1,-1,0,0,1,-1,-1,1};
+int directions_y []= {0,0,1,-1,1,1,-1,-1};
+
+bool IsInGrid (int i, int j,const std::vector<std::vector<char> > &grid){
+    if (i>=0 && i<grid.size()){
```
flerdacodeu: This entire expression can actually be a single `return`-statement.
Sometimes that might be at the cost of readability, but in this case, I think it will be reasonably readable.
If you want to make it more readable, you can also have a helper function `InRange(value, min_value, max_value)` that can help make your intention clearer.

laiaaa0/assignment3/q1.cpp
```diff
@@ -0,0 +1,51 @@
...
+      if (j>=0 && j<grid[i].size()) return true;
+    }
+    return false;
+}
+
+void FindAllWordsRecursive (int x, int y, Dictionary * d, std::vector<std::vector<char> > &grid, std::vector<std::vector<bool> > &visited,  std::string current_word,std::vector<std::string> &found_words){
```
flerdacodeu: Generally, it is good practice to pass anything that cannot be NULL and should not be modified as a "const reference", e.g.:

```c++
  const Dictionary& d
```

This means that you do not have to worry about NULL checks, and you cannot call a "non-const" method on the object by mistake. This will require you to make sure that `Dictionary` has the right `const` methods, e.g.:

```c++
  bool IsPrefix(std::string prefix) const;
```

in your class declaration and definition.

https://google.github.io/styleguide/cppguide.html#Reference_Arguments

laiaaa0/assignment3/TestingFunctions.cpp
```diff
@@ -0,0 +1,104 @@
...
+  EXPECT_FALSE(d->IsWord("RAT"));
+  EXPECT_FALSE(d->IsWord("ca"));
+
+}
+void test_IsPrefixInDictionary (Dictionary * d){
+  EXPECT_TRUE(d->IsPrefix("c"));
+  EXPECT_TRUE(d->IsPrefix("ca"));
+  EXPECT_TRUE(d->IsPrefix("cat"));
+  EXPECT_TRUE(d->IsPrefix("card"));
+  EXPECT_FALSE(d->IsPrefix(""));
```
flerdacodeu: This is not obvious from the intuitive definition of prefix.
But it is a fine choice to make. In that case, however, I would document this where `IsPrefix()` is defined.

laiaaa0/assignment3/q1.cpp
flerdacodeu: Style: split this line as it is very long at the moment, making it hard to read.
The best way to split it is probably after one of the commas.

flerdacodeu: Similar to the "const reference" convention, a common practice is to make sure that anything that you can modify is passed as a "non-const pointer". This help make it clearer what it can and cannot be modified.

laiaaa0/assignment3/q1.cpp
```diff
@@ -0,0 +1,51 @@
...
+  current_word = current_word+grid[x][y];
+  visited[x][y] = true;
+  if (d->IsWord(current_word)){
+    found_words.push_back(current_word);
+  }
+  if (d->IsPrefix(current_word)){
```
flerdacodeu: Optional: sometimes adding a lot of nesting (within `if`- or `for`-statements) makes the code harder to read.
In this case you could do the following:

```c++
  // No word can be formed using this prefix, no need to search any further.
  if (!d->IsPrefix(current_word)) return;

   for (int i = 0; i<8; ++i){
     if (IsInGrid(x+directions_x[i],y+directions_y[i],grid) && !visited[x+directions_x[i]][y+directions_y[i]]){
       FindAllWordsRecursive(x+directions_x[i],y+directions_y[i],d,grid,visited,current_word,found_words);
     }
   }
```

laiaaa0/assignment3/q1.cpp
```diff
@@ -0,0 +1,51 @@
...
+      for (int i = 0; i<8; ++i){
+        if (IsInGrid(x+directions_x[i],y+directions_y[i],grid) && !visited[x+directions_x[i]][y+directions_y[i]]){
```
flerdacodeu: We could simplify the code by introducing some temporary variable. E.g.:

```c++
     int next_x = x + directions_x[i];
     int next_y = y + directions_y[i];
     if (IsInGrid(next_x, next_y, grid) && !visited[next_x][next_y]){
       FindAllWordsRecursive(next_x, next_y, d, grid, visited, current_word, found_words);
     }
```

laiaaa0/assignment3/q1.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+bool IsInGrid (int i, int j,const std::vector<std::vector<char> > &grid){
+    if (i>=0 && i<grid.size()){
+      if (j>=0 && j<grid[i].size()) return true;
+    }
+    return false;
+}
+
+void FindAllWordsRecursive (int x, int y, Dictionary * d, std::vector<std::vector<char> > &grid, std::vector<std::vector<bool> > &visited,  std::string current_word,std::vector<std::string> &found_words){
+  current_word = current_word+grid[x][y];
```
flerdacodeu: Optional: You could use a slightly more compact:

```c++
  current_word += grid[x][y];
```

laiaaa0/assignment3/q1.cpp
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+      }
+  }
+
+
+}
+std::vector<std::string> FindAllWords (Dictionary *d, std::vector<std::vector<char> > &grid){
+  int n = grid.size();
+  int m;
```
flerdacodeu: Define this variable within the loop, where you actually need it.

laiaaa0/assignment3/TestingFunctions.cpp
```diff
@@ -0,0 +1,104 @@
...
+void test_example(){
+  std::vector<std::vector<char> > letter_grid(2, std::vector<char> (3,'.'));
+  CreateSampleGrid(letter_grid);
+
+  std::vector<std::string> v{"car","card","cart","cat"};
+  Dictionary * d = new Dictionary(v);
+  std::vector<std::string> result;
+  result = FindAllWords(d,letter_grid);
+
+  EXPECT_EQ(result.size(),3);
```
flerdacodeu: Can we also check that the words we found are correct?

laiaaa0/assignment3/Dictionary.h
```diff
@@ -0,0 +1,22 @@
...
+    bool IsWord (const std::string & query_word);
+    bool IsPrefix (const std::string & query_prefix);
+    bool InsertWord(const std::string & word);
+    bool RemoveWord(const std::string & word);
+
+
+
+};
+
+std::vector<std::string> FindAllWords (Dictionary *d, std::vector<std::vector<char> > &grid);
```
flerdacodeu: Optional:

This should probably be in a separate `q1.h` file, since the implementation is in `q1.cpp`.
You might also want to split the tests into `dictionary_test.cpp` and `q1_test.cpp`.

laiaaa0/assignment3/Dictionary.cpp
```diff
@@ -1,47 +1,45 @@
 #include "Dictionary.h"
 
-bool Dictionary::IsPrefixOfWord(std::string prefix, std::string word){
-  if (prefix.size()>word.size()) return false;
-  else {
-    for (int i=0; i<prefix.size(); ++i){
-      if (word[i]!=prefix[i]) return false;
-    }
-    return true;
+bool Dictionary::IsPrefixOfWord(std::string prefix, std::string word) const{
```
ssancho2: As Flavio mentioned, string has a compare function that can be used to simplify this, a lot :)
You can basically do:
`return word.compare(0, prefix.size(), prefix) == 0;`

laiaaa0/assignment3/Dictionary.cpp
```diff
@@ -0,0 +1,70 @@
...
+      if (word[i]!=prefix[i]) return false;
+    }
+    return true;
+  }
+}
+
+
+Dictionary::Dictionary(const std::vector<std::string> &word_list){
+  for (int i = 0; i<word_list.size(); ++i){
+    std::string current_word = word_list[i];
```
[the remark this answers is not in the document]
ssancho2: Optionally, if you think it improves readability to give it a name, you can use:
`const std::string &current_word = word_list[i];`

laiaaa0/assignment3/Dictionary.cpp
```diff
@@ -1,47 +1,45 @@
...
 }
-bool Dictionary::IsWord (const std::string & query_word){
-  if (this->dictionary_words_.find(query_word) == this->dictionary_words_.end()) return false;
-  else return (this->dictionary_words_[query_word]);
+
+bool Dictionary::IsWord(const std::string & query_word) const {
+  auto query_iterator =this->dictionary_words_.find(query_word);
+  if (query_iterator == this->dictionary_words_.end()) return false;
+
+  return query_iterator->second;
```
ssancho2: Can this actually be false?

laiaaa0/assignment3/Dictionary.cpp
```diff
@@ -1,47 +1,45 @@
...
-  else {
-    if (this->dictionary_prefixes_[query_prefix]>0) return true;
-    else return false;
-  }
+bool Dictionary::IsPrefix(const std::string & query_prefix) const {
+  auto query_iterator = this->dictionary_prefixes_.find(query_prefix);
+  if (query_iterator == this->dictionary_prefixes_.end()) return false;
 
+  if (query_iterator->second > 0) return true;
+  return false;
```
ssancho2:
```c++
if (query_iterator->second > 0) return true;
return false;
```
could be replaced with:
```c++
return query_iterator->second > 0;
```
or 
```c++
 if (query_iterator == this->dictionary_prefixes_.end()) return false;
 if (query_iterator->second > 0) return true;
 return false;
```
With: 
```c++
return query_iterator != this->dictionary_prefixes_.end() && query_iterator->second > 0;
```
But then readability starts to suffer.

laiaaa0/assignment3/Dictionary.h
```diff
@@ -1,22 +1,27 @@
...
+#include <set>
+
+#ifndef DICTIONARY_H
+#define DICTIONARY_H
+
 class Dictionary{
-  private:
 
+  private:
     std::map<std::string, bool> dictionary_words_;
```
ssancho2: Shouldn't this actually be a `std::set<std::string>` ?
I you didn't not had RemoveWord dictionary_prefixes_ could also be a set.

laiaaa0/assignment3/TestingFunctions.cpp
```diff
@@ -11,94 +10,70 @@ void CreateSampleGrid(std::vector<std::vector<char> > &grid){
...
-  EXPECT_FALSE(d->IsPrefix("card"));
-
-  EXPECT_TRUE(d->RemoveWord("rat"));
-  EXPECT_FALSE(d->IsPrefix("r"));
-  EXPECT_FALSE(d->IsPrefix("ra"));
-  EXPECT_FALSE(d->IsPrefix("rat"));
-
-
 
+void test_IsPrefixAfterRemoval (Dictionary & d){
```
ssancho2: Ideally, tests have no parameters, they are encapsulated and there is no state going in/out.
This makes them easier to read and it should be possible to execute each test in isolation (imagine you are fixing a failing test and all tests take a long time, you will want to run only the failing test while you iterate, and all again afterwards!!).

PR #23, opened 2018-07-01
MennaMohie-4

MennaMohie:


MennaMohie/assignment4/Solution.cpp
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	bool validCol(const int &col)
+	{
+		return col >= 0 && col < cols;
+	}
+
+public:
+
+	Counter(const vector<vector<bool>> g):grid(g), rows(g.size()), cols(g[0].size()), count(0){}
```
ssancho2: Minor detail, rows of the vector may have different sizes.

MennaMohie/assignment4/Solution.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+	void removeIsland(const int currentRow, const int currentCol)
+	{
+		grid[currentRow][currentCol] = false;
+
+		int updatedRow, updatedCol;
+		//iterating over deltaX and deltaY to move up,left,right,down
+		for (int i : {0, 1, 2, 3})
```
ssancho2: Interesting :)

MennaMohie/assignment4/Solution.cpp
```diff
@@ -0,0 +1,115 @@
...
+	bool validCol(const int &col)
+	{
+		return col >= 0 && col < cols;
+	}
+
+public:
+
+	Counter(const vector<vector<bool>> g):grid(g), rows(g.size()), cols(g[0].size()), count(0){}
+
+	void removeIsland(const int currentRow, const int currentCol)
```
ssancho2: `const` here is not needed.

MennaMohie/assignment4/Solution.cpp
```diff
@@ -0,0 +1,115 @@
...
+using namespace std;
+
+class Counter
+{
+	int count, rows, cols;
+	vector<vector<bool>> grid;
+	int deltaX[4] = { -1,0,0,1 };
+	int deltaY[4] = { 0,-1,1,0 };
+
+	bool validRow(const int &row)
```
ssancho2: Some comment on style since I don't have much else :)
It may be more legible to put the '{' in the previous line, so more stuff fits on the screen at once.

MennaMohie/assignment4/Solution.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+public:
+
+	Counter(const vector<vector<bool>> g):grid(g), rows(g.size()), cols(g[0].size()), count(0){}
+
+	void removeIsland(const int currentRow, const int currentCol)
+	{
+		grid[currentRow][currentCol] = false;
+
+		int updatedRow, updatedCol;
```
ssancho2: Better to declare them when you are setting them.

PR #24, opened 2018-07-03
Laiaaa0 4

laiaaa0:


laiaaa0/assignment4/Map.h
```diff
@@ -0,0 +1,35 @@
+#include <vector>
+#include <iostream>
+
+#ifndef MAP_H
```
ssancho2: Minor thing: this usually goes before everything else. It doesn't really mater because standard headers have their own guards and will not be included twice tough.

laiaaa0/assignment4/TestingFunctions.cpp
```diff
@@ -0,0 +1,80 @@
...
+    tiles[1][0] = 'T';
+    tiles[0][3] = 'T';
+    tiles[2][2] = 'T';
+    tiles[3][2] = 'T';
+    Map map(tiles);
+    EXPECT_EQ(3,map.FindNumberIslands());
+}
+
+void test_ExampleTwoIslands(){
+    std::vector<std::vector<char > > tiles(5,std::vector<char> (5,'T'));
```
ssancho2: Won't this be all 'T' s?

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
...
+    this->num_columns_ = tiles[0].size();
+  }
+  else {
+    this->num_columns_ = 0;
+  }
+  this-> map_tiles_ = tiles;
+
+}
+
+Map::Map(const std::string & tiles_str) {
```
ssancho2: Nice! :)

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
...
+Map::Map(const std::vector<std::vector<char> > & tiles) {
+  this->num_rows_ = tiles.size();
+  if (num_rows_ > 0){
+    this->num_columns_ = tiles[0].size();
+  }
+  else {
+    this->num_columns_ = 0;
+  }
+  this-> map_tiles_ = tiles;
+
```
ssancho2: empty line not needed.

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
+#include "Map.h"
+
+int directions_x [] = {0,0,-1,1};
+int directions_y [] = {1,-1,0,0};
+
+Map::Map(const std::vector<std::vector<char> > & tiles) {
+  this->num_rows_ = tiles.size();
+  if (num_rows_ > 0){
+    this->num_columns_ = tiles[0].size();
```
ssancho2: Minor detail: rows may have different sizes.

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
...
+    this->num_columns_ = 0;
+  }
+  this-> map_tiles_ = tiles;
+
+}
+
+Map::Map(const std::string & tiles_str) {
+  this->map_tiles_.clear();
+  std::vector<char> current_row;
+  for (int i = 0; i<tiles_str.size(); ++i){
```
ssancho2: You can do:
`for (char c : tiles_str) { ... }`

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
...
+        this->map_tiles_.push_back(current_row);
+        current_row.clear();
+      }
+      else {
+        if (tiles_str[i]==ISLAND_T || tiles_str[i]==WATER_T)  {
+          current_row.push_back(tiles_str[i]);
+        }
+      }
+  }
+  this->num_rows_ = this->map_tiles_.size();
```
ssancho2: I've seen this code to initialize `num_rows_` & `num_columns_` before :) 
Actually, it may be easier to get `map_tiles_.size()` & `map_tiles_[row].size()` when needed?

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
...
+bool Map::InRange(int value, int min, int max) {
+    return value>=min && value<=max;
+}
+
+bool Map::IsInMap(int x, int y) {
+    return InRange(x, 0, this->num_rows_-1) && InRange(y, 0, this->num_columns_-1);
+}
+
+
+bool Map::ExpandNode(int x, int y, std::vector<std::vector<bool> > & visited) {
```
ssancho2: Minor thing: pointers are a bit better than references for things you will modify, since it makes it clear to the client.
If you see:
`ExpandNode(x, y, visited)`
you don't know what will happen with visited, but if you see:
`ExpandNode(x, y, &visited)`
is clear that visited may be modified.

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
...
+    if (!IsInMap (x,y)) return false;
```
ssancho2: Minor thing: the ifs can be collapsed.

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
...
+    if (visited[x][y]) return false;
+    visited[x][y] = true;
+    if (this->map_tiles_[x][y] == WATER_T){
+        return false;
+    }
+    else if (this->map_tiles_[x][y] == ISLAND_T){
```
ssancho2: Either the else and if are not needed, or you need a extra else to handle the case of a tile that is neither water nor land.

laiaaa0/assignment4/Map.cpp
```diff
@@ -0,0 +1,106 @@
...
+        for (int i = 0; i < 4; ++i){
+                int next_x = x + directions_x[i];
+                int next_y = y + directions_y[i];
+                ExpandNode(next_x, next_y, visited);
+        }
+        return true;
+    }
+}
+
+std::pair<int,int> Map::GetNextUnvisited(int x, int y, const std::vector<std::vector<bool> > & visited) {
```
ssancho2: It's a bit confusing (at least for me), using x for rows and y for columns, since Cartesian graphs are usually the other way around.

Also, do you need this function? I think it's easier to iterate over all cells on `FindNumberIslands`

laiaaa0/assignment4/TestingFunctions.cpp
```diff
@@ -0,0 +1,80 @@
...
+    for (int i = 0; i < 5; ++i){
+      for (int j = 0; j < 5; ++j){
+        tiles[i][j] = 'T';
+      }
+    }
+    Map map(tiles);
+    EXPECT_EQ(map.FindNumberIslands(),1);
+}
+void test_CreateMapFromStr(){
+  std::string map_str = "FTFT-"
```
ssancho2: Nice! :)

laiaaa0/assignment4/Map.cpp
laiaaa0: I think the code is more legible if separating the functions, it made more sense in my head! But maybe not...

laiaaa0/assignment4/Map.cpp
```diff
@@ -77,18 +61,22 @@ std::pair<int,int> Map::GetNextUnvisited(int x, int y, const std::vector<std::ve
...
 int Map::FindNumberIslands() {
-    std::vector<std::vector<bool> > visited(this->num_rows_, std::vector<bool> (this->num_columns_,false));
+    int num_columns = 0;
+    if (this->map_tiles_.size()>0) num_columns = this->map_tiles_[0].size();
+    std::vector<std::vector<bool> > visited(this->map_tiles_.size(), std::vector<bool> (num_columns,false));
     int x = 0, y = 0;
     int num_islands = 0;
-    while (x<this->num_rows_ && y < this->num_columns_){
-        if (this->ExpandNode (x,y,visited)) num_islands++;
+    while (x<this->map_tiles_.size() && y < this->map_tiles_[x].size()){
```
ssancho2: Checking for y in range is enough, and saves you setting `next_unvisited.second` inside `Map::GetNextUnvisited()`

laiaaa0/assignment4/Map.cpp
```diff
@@ -77,18 +61,22 @@ std::pair<int,int> Map::GetNextUnvisited(int x, int y, const std::vector<std::ve
...
+      next_unvisited.second = this->map_tiles_[0].size();
+    }
     return next_unvisited;
 }
 
 int Map::FindNumberIslands() {
-    std::vector<std::vector<bool> > visited(this->num_rows_, std::vector<bool> (this->num_columns_,false));
+    int num_columns = 0;
+    if (this->map_tiles_.size()>0) num_columns = this->map_tiles_[0].size();
+    std::vector<std::vector<bool> > visited(this->map_tiles_.size(), std::vector<bool> (num_columns,false));
```
ssancho2: To handle Jagged matrices, this need to be replaced by a loop so each row on visited has the same size as the corresponding row on map_tiles_

PR #25, opened 2018-07-03
MayssaraOmar-4 add assignment 4 solution 

MayssaraOmar:


MayssaraOmar/assignment4/Counting islands.cpp
```diff
@@ -0,0 +1,125 @@
+#include "test.h"
+#include <iostream>
+#include <vector>
+using namespace std;
+
+static void markIsland(
+	const vector<vector<bool>> &grid,
```
ssancho2: Indentation is a bit off here, parameters should be a little more Indented than the code.
Also, having one in the first line may be more clear.

MayssaraOmar/assignment4/Counting islands.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// tests
+void emptyGrid_test() {
+	vector<vector<bool>> grid;
+	EXPECT_EQ(0, getIslandsCount(grid));
+}
+
+void landGrid_test() {
+	vector<vector<bool>> grid = {{true, true, true},
```
ssancho2: Minor comment, having a unction to convert `string` -> `vector<vector<bool>>` or `vector<string>` -> `vector<vector<bool>>` count make the test cases more readable.

MayssaraOmar/assignment4/Counting islands.cpp
```diff
@@ -0,0 +1,125 @@
...
+	vector<vector<bool>> *visited) {
+	// boundary check
+	if (row < 0 || row >= grid.size() || col < 0 || col >= grid[row].size()) {
+		return;
+	}
+	if ((*visited)[row][col] || !grid[row][col]) {
+		return;
+	}
+	(*visited)[row][col] = true;
+	markIsland(grid, row - 1, col, visited); // up
```
ssancho2: Maybe having a loop (or two) makes this simpler. Not sure tough, looks good as it is...

PR #26, opened 2018-07-03
JoodHajeer-4

JoodHajeer:


JoodHajeer/assignment4/Source.cpp
```diff
@@ -0,0 +1,55 @@
+#include <iostream>
+#include <stdio.h>  
+#include <string>
+#include <queue>
+#include <set>
+#include <stdlib.h>
+#include"codeu_test_lib.h"
+using namespace std;
+
+void DFS(int curRow, int curCol, int numberOfRows, int numberOfColumns, vector<vector<bool> > & tilesMap) {
```
ssancho2: Minor thing: pointers are a bit better than references for things you will modify, since it makes it clear to the client.
If you see:
`DFS(..., tilesMap)`
you don't know what will happen with tilesMap, but if you see:
`DFS(..., &tilesMap)`
is clear that tilesMap may be modified.

JoodHajeer/assignment4/Source.cpp
```diff
@@ -0,0 +1,55 @@
...
+	int dc[] = {1,-1,0,0};
+	for (int i=0;i<4;i++){
+		int newRow = curRow + dr[i];
+		int newCol = curCol + dc[i];
+		if (newRow < 0 || newRow >= numberOfRows || newCol < 0 || newCol >= numberOfColumns || !tilesMap[newRow][newCol]) // skip tiles that are out on bound or water or already visited 
+			continue;
+		DFS(newRow, newCol, numberOfRows, numberOfColumns, tilesMap);
+	}
+}
+int countIslands(int numberOfRows, int numberOfColumns, vector<vector<bool> > tilesMap) {
```
ssancho2: The copy of tilesMap is intentional, right? Nice :)
Note: numberOfColumns & numberOfRows are redundant, they can be extracted from tilesMap. Also, minor thing, but the rows of tilesMap may have different sizes.

JoodHajeer/assignment4/Source.cpp
```diff
@@ -0,0 +1,55 @@
...
+	for (int i = 0; i < numberOfRows; i++)
+		for (int j = 0; j < numberOfColumns; j++)
+			if (tilesMap[i][j]) { //found an island
+				DFS(i, j, numberOfRows, numberOfColumns,tilesMap); //color all tiles of the current island
+				++numberOfIslands;
+			}
+	return numberOfIslands;
+}
+
+void testCountIslands_emptyTilesMap(int numberOfRows, int numberOfColumns, const vector<vector<bool> > tilesMap) {
```
ssancho2: tilesMap can be a `const &`, to avoid the copy.

JoodHajeer/assignment4/Source.cpp
```diff
@@ -0,0 +1,55 @@
...
+void testCountIslands_singleCol(int numberOfRows, int numberOfColumns, const vector<vector<bool> > tilesMap) {
+	int answer = countIslands(numberOfRows, numberOfColumns, tilesMap);
+	EXPECT_EQ(2, answer);
+}
+void testCountIslands_givenCase(int numberOfRows, int numberOfColumns, const vector<vector<bool> > tilesMap) {
+	int answer = countIslands(numberOfRows, numberOfColumns, tilesMap);
+	EXPECT_EQ(3, answer);
+}
+int main() {
+	testCountIslands_emptyTilesMap(0, 0, {});
```
ssancho2: Having the parameters here and the expected answer inside the function is kind of strange. Either:
Also pass the expected value (and then all test* functions become one)
(preferred) Pass no parameters the the test* function, and have the function do all, for example:
```c++
void testCountIslands_givenCase() {
	int answer = countIslands(4, 4, { {0,1,0,1},{1,1,0,0},{0,0,1,0},{0,0,1,0} });
	EXPECT_EQ(3, answer);
}
```

PR #27, opened 2018-07-03
Unit tests to be done

Lacrit:


Lacrit/assignment4/islands.cpp
```diff
@@ -0,0 +1,68 @@
+#include <stdio.h>
+#include <string.h>
+#include <stdbool.h>
+ 
+const int rows = 5;
+const int cols = 5;
+
+// A function to check if a given cell (row, col) can be included in DFS
+int isValid(int mat[rows][cols], int row, int col, bool visited[rows][cols])
```
ssancho2: It will be better if the code can have matrices of different sizes. This could be done having three parameters: `int **mat, int num_rows, int num_cols`, or preferably using a `vector<vector<int>>`.
For example, this function could become:
```c++
bool isValid(const vector<vector<bool>> &mat, int row, int col, const vector<vector<bool>> &visited) 
{
    // row number is in range, column number is in range and value is 1 
    // and not yet visited
    return (row >= 0) && (row < mat.size()) &&     
           (col >= 0) && (col < mat[row].size()) &&      
           (mat[row][col] && !visited[row][col]); 
}
```
Marking them as a `const &` is not strictly needed, but without them the compiler will make a copy of the matrices every time we call this functions, which is a bit slow.

Lacrit/assignment4/islands.cpp
```diff
@@ -0,0 +1,68 @@
...
+ 
+    // Recur for all connected neighbours
+    for (int k = 0; k < 8; ++k)
+        if (isValid(mat, row + rowInd[k], col + colInd[k], visited) )
+            DFS(mat, row + rowInd[k], col + colInd[k], visited);
+}
+ 
+// The main function that returns count of islands in a given boolean
+// 2D matrix
+int countIslands(int mat[rows][cols])
```
ssancho2: Minor detail, but mat should be a **bool** 2D matrix, as the comment says. :)

Lacrit/assignment4/islands.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // These arrays are used to get row and column numbers of 8 neighbours 
+    // of a given cell
+    static int rowInd[] = {-1, -1, -1,  0, 0,  1, 1, 1};
+    static int colInd[] = {-1,  0,  1, -1, 1, -1, 0, 1};
+ 
+    // Mark this cell as visited
+    visited[row][col] = true;
+ 
+    // Recur for all connected neighbours
+    for (int k = 0; k < 8; ++k)
```
ssancho2: It may be slightly easier to have two cycles:
```c++
for (int dx : {-1, 0, 1}) {
  for (int dy : {-1, 0, 1}) {
    ...
  }
}
```
or using `for (int dx = -1; dx <= 1; dx++) { ... } `

Note that isValid will return false when dx==0 & dy==0, since that cell is already painted as visited.

But since we need to iterate over just 4 neighbours is probably easier as you did :)

Lacrit/assignment4/islands.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // and not yet visited
+    return (row >= 0) && (row < rows) &&     
+           (col >= 0) && (col < cols) &&      
+           (mat[row][col] && !visited[row][col]); 
+}
+ 
+// check all 8 neighbours
+void DFS(int mat[rows][cols], int row, int col, bool visited[rows][cols])
+{
+    // These arrays are used to get row and column numbers of 8 neighbours
```
ssancho2: Note that the problem statement specifies that we do **not** take diagonal connections.

Lacrit/assignment4/Islands.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "Islands.h"
+
+int rowInd[] = {-1, 1, 0, 0};
+int colInd[] = {0, 0, 1, -1};
+
+Islands::Islands(std::vector<std::vector<bool>> grid) 
+{
+	this->grid = grid;
+	this->rows = grid.size();
+	this->cols = grid[0].size();
```
ssancho2: Minor comment, grid rows can potentially be of different size, ideally you want to either check that they are all of the same size or handle the case where they are not.

Lacrit/assignment4/Islands.cpp
```diff
@@ -0,0 +1,55 @@
...
+};
+
+// A function to check if a given cell (row, col) can be included in DFS
+bool Islands::isValid(const std::vector<std::vector<bool>> &mat, int row, int col, int rows, int cols, const std::vector<std::vector<bool>> &visited)
```
ssancho2: You don't need to pass cols & rows, they are fields of the class, right?

Lacrit/assignment4/Islands.cpp
```diff
@@ -0,0 +1,55 @@
...
+    for (int k = 0; k < 4; ++k)
+        if (isValid(mat, row + rowInd[k], col + colInd[k], rows, cols, visited))
+            DFS(mat, row + rowInd[k], col + colInd[k], rows, cols, visited);
+}
+ 
+// The main function that returns the number of islands in a given boolean 2D matrix
+int Islands::countIslands(const std::vector<std::vector<bool>> &mat, int rows, int cols)
+{
+    // Make a bool array to mark visited cells.
+	// Initially all cells are unvisited
```
ssancho2: Indentation is a bit inconsistent here.

Lacrit/assignment4/Islands.h
```diff
@@ -0,0 +1,15 @@
+#include <vector>
+
+class Islands 
+{
+	private: 
+		std::vector<std::vector<bool>> grid; 
+	public: 
+		int rows;
```
ssancho2: Why rows & cols are public?

Lacrit/assignment4/Islands.h
```diff
@@ -0,0 +1,15 @@
...
+		int cols;
+		Islands(std::vector<std::vector<bool>> grid);
+		bool isValid(const std::vector<std::vector<bool>> &mat, int row, int col, int rows, int cols, const std::vector<std::vector<bool>> &visited);
```
ssancho2: `isValid()` & `DFS()` could also be private. They are not useful for a user, right? They are part of the implementation of `countIslands()`

Lacrit/assignment4/Unit_tests.cpp
```diff
@@ -0,0 +1,88 @@
...
+void test_emptyGrid()
+{
+	std::vector<std::vector<bool>> grid
+	{};
+	Islands i(grid);
+    EXPECT_EQ(0, i.countIslands(grid, i.rows, i.cols));	
+
+}
+
+void tets_noWater()
```
ssancho2: test -> test :)
Also, this is all water, that's why the result is 0.

Lacrit/assignment4/Unit_tests.cpp
```diff
@@ -0,0 +1,88 @@
...
+        {0, 0, 0, 0},
+        {0, 0, 0, 0},
+        {0, 0, 0, 0}
+	};
+	Islands i(grid);
+    EXPECT_EQ(0, i.countIslands(grid, i.rows, i.cols));	
+
+}
+
+void test_allWater()
```
ssancho2: And this is all land.

PR #28, opened 2018-07-13
MennaMohie-5

MennaMohie:


group/assignment5/Solution.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+		//filling the visited map
+		unordered_set<char>::iterator it;
+		for (it = allChars.begin(); it != allChars.end(); it++)
+		{
+			visited[*it] = false;
+		}
+
+		//iterating over all chars
+		for (it = allChars.begin(); it!=allChars.end(); it++)
```
marahgh: Question:
say allChars ={a,b} and b<a ,i'ts not clear to me how you insure to visit b before a

marahgh: cool

PR #30, opened 2018-07-14
Prolly needs refactoring

Lacrit:


Lacrit/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,42 @@
+#include <fstream>
+#include <iostream>
+#include <unordered_set>
+#include <string>
+
+
+class Dictionary {
+public:
+	Dictionary();
```
ssancho2: Please try to keep indentation consistent.

Lacrit/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,42 @@
...
+    Dictionary(std::string);
```
ssancho2: Even if nod needed by the compiler, putting names for the parameters helps readers understand what parameters are.

Lacrit/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,42 @@
...
+    ~Dictionary();
+	void Fill(std::string);
+	bool isPrefix(const std::string&); 
+
+    std::unordered_set<std::string> words;
```
ssancho2: Those probably don't need to be public.

Lacrit/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,42 @@
...
+{}
+// Load the word dictionary and prefixes dictionary from a given file
+void Dictionary::Fill(std::string path) 
+{
+    std::string line;
+    std::ifstream file(path);
+    if (file.is_open()) {
+        while (std::getline(file, line)) {
+                // Add to word dictionary
+                this->words.insert(line);
```
ssancho2: `this->` is not needed.

Lacrit/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,42 @@
...
+	bool isPrefix(const std::string&); 
+
+    std::unordered_set<std::string> words;
+    std::unordered_set<std::string> prefixes;
+	std::string path;
+};
+
+Dictionary::Dictionary() {};
+Dictionary::Dictionary(std::string path):
+	path(path)
```
ssancho2: It helps readability having different naming conventions for fields and parameters. On the Google Style for example, we end fields with '_'.

Lacrit/assignment3/Main.hpp
```diff
@@ -0,0 +1,125 @@
...
+	this->dict = dict;
+	int row = 0; 
+	int col = 0;
+
+	
+  for (int i = 0; i < rows; i++) {
+	  row = i / rows;
+	  for ( int j = 0; j < cols; j++) {
+        col = j % cols;
+        board.push_back(cell(row, col));
```
ssancho2: (*)
I think here `row` is 0 and `col` is `j`.
You can use `i` and `j` directly, right?

Lacrit/assignment3/Main.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+  //get all neighbours of a cell
+  std::vector<int> shift {-1, 0, 1};
+    for (cell &sq : board) {
+        for (int row_shift : shift) {
+            for (int col_shift : shift) {
+                row = sq.row + row_shift;
+                col = sq.col + col_shift;
+                if (row >= 0 & row < rows & col >= 0 & col < cols & !(row_shift == 0 & col_shift == 0)) {
+                    sq.neighbours.push_back(row * rows + col);
```
ssancho2: (*)
`row * rows + col` should be either:
`row + rows * col` or `row * cols + col`
One way to think about it is (for the first one):
When col = 0, this goes from 0 -> rows -1, so the next available number when col = 1 is rows.
In that case, if `n = row + rows * col`
You can extract them back  (using that row < rows, and they are both positive integers):
`n / rows -> (row + rows * col) / rows -> row / rows + (rows * col) / rows -> 0 + col -> col`
`n % rows -> (row + rows * col) % rows -> row % rows + (rows * col) % rows -> row + 0 -> row`

Lacrit/assignment3/Main.hpp
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  //get all neighbours of a cell
+  std::vector<int> shift {-1, 0, 1};
+    for (cell &sq : board) {
+        for (int row_shift : shift) {
+            for (int col_shift : shift) {
+                row = sq.row + row_shift;
+                col = sq.col + col_shift;
+                if (row >= 0 & row < rows & col >= 0 & col < cols & !(row_shift == 0 & col_shift == 0)) {
```
ssancho2: it works, but you probably should use logical and `&&` instead of bitwise and `&`

Lacrit/assignment3/Main.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+// Find all words starting at a given position
+void Main::Words(int position, std::string string, std::unordered_set<int> visited) {
+
+  cell &sq = board[position];
+    string += sq.value;
+    visited.insert(position);
+
+    // If the string is a word, add it to the found words
+    if (dict.words.find(string) != dict.words.end()) {
```
ssancho2: This could be a function `isWord()` in the dictionary class, and the code bellow `(dict.prefixes.find(string) != dict.prefixes.end())` ==>  `isPrefix()`

Lacrit/assignment3/Main.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+// prolly not the best solution to store it like that
+int rows = 2;
+int cols = 3;
+
+class Main {
+private:
+ 	struct cell {
+        std::string value;
+        int row;
```
ssancho2: It's ok to store them, but they are technically not needed.
As I explain in more detail in a comment above, you can from int index to coordinates and back 'easily'
Using a int to refer to the index of a cell in the board vector is ok, but a few comments will help :)

PR #34, opened 2018-07-20
Group solution

marahgh:


group/assignment5/AlphabetFinder.h
```diff
@@ -0,0 +1,47 @@
...
+
+    // A helping function of the topologicalSort.
+    void visit(char c);
+
+   public:
+    PrecedenceGraph(vector<string> words);
+
+    // makes a topological Sort on the graph.
+    // returns a stack of the sorted graph elements.
+    stack<char> topologicalSort();
```
MayssaraOmar: shouldn't we make this function private and keep the `getAlphabet()` function only public?

group/assignment5/Testing.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <algorithm>
+#include <vector>
+#include <stack>
+#include <map>
+#include <unordered_map>
+#include <unordered_set>
+#include "TestingMacros.h"
+
+void givenTest() {
+  AlphabetFinder AF({"ART", "RAT", "CAT", "CAR"});
```
MayssaraOmar: Just a side note that if we decide to call the dictionary by const reference, then using temporary vectors as in `AlphabetFinder AF({"ART", "RAT", "CAT", "CAR"});` will cause an error and we'll need to do something like this: 
```
vector<string> dictionary {"ART", "RAT", "CAT", "CAR"};
AlphabetFinder AF(dictionary);
```

group/assignment5/AlphabetFinder.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+AlphabetFinder::AlphabetFinder(vector<string> dictionary)
+    : rules(PrecedenceGraph(dictionary)) {}
+
+vector<char> AlphabetFinder::getAlphabet() {
+  stack<char> alphabetStack = rules.topologicalSort();
+
+  // filling the alphabet vector
+  while (!alphabetStack.empty()) {
```
ssancho2: The reversing should be probably part of the topologicalSort

